PR #2, opened 2018-01-10
start generating grpc services, add go/types

mvdan:


mvdan:
Added more stuff to the PR. Most notably, we now use the constant folding information from go/types, and we support loading .proto dependencies.

PR #4, opened 2018-01-16
first working version of gunk packages importing other gunk packages

mvdan:


mvdan:
CI is now working too.

PR #5, opened 2018-01-17
First fully working version, with generated Go code that builds

mvdan:
Including gunk packages importing other gunk packages.

See the commits for more details, as this was trickier than it may initially seem.


mvdan:
Ah, the vendor thingie isn't working as I thought it would.

PR #10, opened 2018-10-22
loader: handle int, bool, array and struct

vishen:
The Gunk example now works.

'protoType' can now return a 'label', this is useful for any fields that
need to be 'repeated', 'optional' or 'required' in proto.

'protoType' handles bool, int, int32, array and struct. Array's are now
possible because a 'label' is returned. For struct's and enum's, we check whether
the 'enum' or 'message' have been parsed and are in a valid gunk file.
This is likely not the most robust way, but works for now.

Added a slightly more useful error message when 'protoc-gen-go' fails
that includes and stdout and stderr output.

Added the command to run the example in the readme.


loader/loader.go
```diff
@@ -479,18 +482,37 @@ func (l *Loader) protoEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, erro
...
 		pkg = l.tpkg
 	}
 	switch x := expr.(type) {
 	case *ast.Ident:
 		switch x.Name {
 		case "string":
-			return desc.FieldDescriptorProto_TYPE_STRING, x.Name
+			return desc.FieldDescriptorProto_TYPE_STRING, 0, x.Name
+		case "int", "int32":
+			return desc.FieldDescriptorProto_TYPE_INT32, 0, x.Name
```
mvdan: This seems fine - it can be a pain to work with `int32` or `int64` in Go code. But I wonder if it would be better to translate `int` as `INT64`. Then again, I guess that if anyone really needs the bits, they should use a 64-bit number for portability.

loader/loader.go
```diff
@@ -479,18 +482,37 @@ func (l *Loader) protoEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, erro
...
+			for _, et := range l.pfile.EnumType {
+				if *et.Name == x.Name {
+					return desc.FieldDescriptorProto_TYPE_ENUM, 0, fullName
+				}
+			}
+			for _, mt := range l.pfile.MessageType {
+				if *mt.Name == x.Name {
+					return desc.FieldDescriptorProto_TYPE_MESSAGE, 0, fullName
+				}
+			}
```
mvdan: Agreed that this can be cleaned up later. Fine to leave as it is for now with TODOs.

loader/loader.go
```diff
@@ -499,8 +521,14 @@ func (l *Loader) protoType(expr ast.Expr, pkg *types.Package) (desc.FieldDescrip
 		}
 		pkg := l.info.ObjectOf(id).(*types.PkgName).Imported()
 		return l.protoType(x.Sel, pkg)
+	case *ast.ArrayType:
```
mvdan: You probably want to allow `[]SomeType`, but error on `[3]SomeType`.

vishen:
How do you add tests? And how do you run against the testdata?

mvdan:
You can run them with `go test` in the root, unless something has broken.

Then `testdata/src` has a bunch of Go packages. Each `.gunk` file has a `.pb.go` file, and the tests check that the generated files match what's on disk.

So to add the tests, simply add some fields somewhere with the types you added support for.

vishen:
Actually, this doesn't handle structs and enums that are defined in other files correctly.

One approach is to parse and load all the top-level types (without loading the struct-body), and then when all the top-level types are loaded, load all of the struct-bodies?

vishen:
This should now handle structs and consts defined in other gunk files in other packages -- @mvdan PTAL.

loader/loader.go
```diff
@@ -479,18 +482,46 @@ func (l *Loader) protoEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, erro
...
+			return desc.FieldDescriptorProto_TYPE_BOOL, 0, x.Name
 		default:
 			fullName := "." + pkg.Path() + "." + x.Name
-			return desc.FieldDescriptorProto_TYPE_ENUM, fullName
+			// Check if the identifier is an already known 'message' or 'enum.
+			// NOTE: This checks all of the parsed gunk packages for the type.
+			// TODO: There is likely a much more robust way to handle this.
+			for _, p := range l.astPkgs[pkg.Path()] {
+				for _, decl := range p.Decls {
+					gd := decl.(*ast.GenDecl)
```
mvdan: This will panic if a package happenst to have a top-level func declaration. Use the `, ok` form, and skip if `!ok`.

loader/loader.go
```diff
@@ -479,18 +482,46 @@ func (l *Loader) protoEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, erro
...
+					for _, spec := range gd.Specs {
+						if ts, ok := spec.(*ast.TypeSpec); ok {
```
mvdan: Early continue on `!ok` to unindent the rest

mvdan:
Also, another minor nit - use `loader:` as the commit prefix, not `loader/loader:`.

vishen:
Done! @mvdan PTAL

loader/loader.go
mvdan: If the old code did `gd, _ := decl.(*ast.GenDecl)` then perhaps, but if you use the non-bool form, the expression will panic instead of giving you nil. Thanks for fixing it :)

vishen: Ah sorry, I was referring to the function that calls this (rather than old code) - https://github.com/gunk/gunk/blob/master/loader/loader.go#L236

vishen:
Yeah, I am having some issues trying to generate the protobufs in the "testdata" folder :(

PR #13, opened 2018-10-24
all: port to a Go module, drop Go 1.9

mvdan:
And bump all dependencies, and re-vendor everything.

Also drop tip, as that's a target that can break. I develop on tip
already, so it's already being tested to some degree.

Finally, disable the tests on CI while we refactor them to work within
the module.


mvdan:
This is the first step towards getting the tests fully working, which will require some work because `go/importer` is obsolete with modules.

mvdan:
Scrapping this, actually. Will try doing everything at once, to not disable the tests.

PR #14, opened 2018-10-29
all: replace go/build with go/packages

### LINKED ISSUE #11: use go/packages
go/loader is now deprecated, and doesn't support Go modules.

mvdan:
go/build is deprecated, since it was designed before Go modules. Its
replacement in most cases, including ours, is go/packages.

Start using it to locate packages, which is the same reason we were
using go/build for. This has two major advantages, besides removing a
deprecated dependency:

* Module support. We no longer depend on GOPATH to resolve imports and
  find other Gunk packages. Moreover, when using a module, missing deps
  will be added to go.mod via the 'go list' invocations.

* Pattern support. go/build did not resolve patterns like ./..., but
  go/packages (again, via 'go list') does.

Fixes #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,7 @@ syntax.
 
 Gunk can be installed in the usual Go fashion:
 
-```sh
-$ go get -u github.com/gunk/gunk
-```
+	go get -u github.com/gunk/gunk
 
 ## Syntax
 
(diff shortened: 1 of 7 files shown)


mvdan:
This is the smallest change I could do without breaking anything. `gunk ./examples/util` still panics with `panic: unknown or mismatched type: type: loader.httpRule, wire type: varint`, which I need to investigate, but otherwise the entire tool should be have the same.

mvdan:
master is already broken on CI, so this is just the same.

PR #16, opened 2018-10-29
CI: fix build via modules

mvdan:
Start using the full power of modules on CI. To do that, we must drop Go
1.10, as it only has minimal support for them.

While at it, drop tip too, as it's a moving target and not very useful
to enforce on CI. I develop on tip, so if an upcoming Go release were to
require changes in Gunk, we'd notice anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
-language: go
-go:
-  - 1.10.x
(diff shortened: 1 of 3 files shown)


mvdan:
The tests aren't reenabled yet, but that will come at a later time.

PR #20, opened 2018-10-30
loader: generate http rules using grpc-gateway

### LINKED ISSUE #17: also generate via protoc-gen-grpc-gateway if any http rules are present in a package
Similar to what we do with protoc-gen-go for all packages currently.

Already spoke to Jon and explained roughly how this would be done.

vishen:
Call out to 'protoc-gen-grpc-gateway'.

Updated 'examples/util/echo.gunk' to remove any paths with trailing
slashes ('/') because grpc-gateway doesn't allow trailing slashes.

Updated the 'interpretTagValue' function to evaluate the entire
expression before creating an 'annotations.HttpRule'. The previous way
meant that if 'Method' wasn't evaluated before 'Path', the created
HttpRule would be incorrect (because 'Path' was assuming 'Method' always
got evaluated first.

Closes #17


go.sum
```diff
@@ -1,12 +1,34 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
mvdan: you ran `go mod tidy`? this is a lot of new sum lines

loader/loader.go
```diff
@@ -398,25 +438,47 @@ func (l *Loader) interpretTagValue(tag string) (*proto.ExtensionDesc, interface{
...
+				// https://github.com/grpc-ecosystem/grpc-gateway/issues/472
 			case "Body":
-				rule.Body = val
+				body = val
 			}
 		}
+		rule := &annotations.HttpRule{
+			Body: body,
+		}
+		switch method {
```
mvdan: I presume you moved this so that a path appearing before a method works as expected

loader/loader.go
```diff
@@ -398,25 +438,47 @@ func (l *Loader) interpretTagValue(tag string) (*proto.ExtensionDesc, interface{
...
-				}
+				path = val
+				// TODO: grpc-gateway doesn't allow paths with a trailing '/', should
+				// we return an error here, because the error from grpc-gateway is very
+				// cryptic and unhelpful?
+				// https://github.com/grpc-ecosystem/grpc-gateway/issues/472
 			case "Body":
-				rule.Body = val
+				body = val
 			}
```
mvdan: while at it, make the default here error too

loader/loader.go
```diff
@@ -127,12 +127,29 @@ func New(dir string, paths ...string) (*Loader, error) {
...
+		return fmt.Errorf("error generating plugin go: %v", err)
+	}
+	if err := l.generatePluginGrpcGateway(*req); err != nil {
+		return fmt.Errorf("error generating plugin grpc-gateway: %v", err)
+	}
+	return nil
+}
+
+func (l *Loader) generatePluginGo(req plugin.CodeGeneratorRequest) error {
+	req.Parameter = proto.String("plugins=grpc")
```
mvdan: can we not leave this in for the request in all generators?

loader/loader.go
```diff
@@ -127,12 +127,29 @@ func New(dir string, paths ...string) (*Loader, error) {
 //
 // Generated files are written to the same directory, next to the source gunk
 // files.
+//
+// It is fine to pass every 'plugin.CodeGeneratorRequest' to every protoc
+// generator unaltered; this is what protoc does when calling out to the
+// generators and the generators should already handle the case where
+// they aren't required.
```
mvdan: I think "where they have nothing to do" is clearer here.

go.sum
vishen: I ran it a few times, but not for the last commit, but running it on this commit is adding more lines as well (I'll commit them now).

loader/loader.go
vishen: Yup, I added some comments about this above.

loader/loader.go
vishen: Not for `grpc-gateway` at least, it fails horribly.

loader/loader.go
vishen: Added.

go.sum
vishen: Ran it again.

PR #23, opened 2018-10-30
loader: make full use of go/types

### LINKED ISSUE #12: don't check builtin types by identifier names
We can't check if one is using a builtin `int` type by checking that the type node is an `*ast.Ident` with `Name == "int"`. That can very easily get confused:

```
type int []string

var NotAnInt int
```

mvdan:
For some reason, we were being silly and inspecting the syntax tree to
translate types from Gunk to protobuf. This is much simpler and robust
with go/types.

Also solves the TODO, since we no longer have to resolve SelectorExpr
nodes ourselves.

Otherwise, nothing is changed. Added more field types to testdata to
ensure that we didn't break a thing.

Fixes #12.


loader/loader.go
```diff
@@ -545,69 +547,38 @@ func (l *Loader) protoEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, erro
...
-		pkg := l.info.ObjectOf(id).(*types.PkgName).Imported()
-		return l.protoType(x.Sel, pkg)
-	case *ast.ArrayType:
-		// Array's arent handled, only slices. Check if this is an array.
-		if x.Len != nil {
-			return 0, 0, ""
+		case *types.Struct:
+			return desc.FieldDescriptorProto_TYPE_MESSAGE, 0, fullName
+		default:
+			fmt.Printf("%T %v\n", typ, typ)
```
vishen: Does leaving this print in help at all?

mvdan: Nope! Just a debugging print. Thanks for spotting.

PR #27, opened 2018-10-31
loader: support time.Time and time.Duration fields

### LINKED ISSUE #26: support timestamps and durations
See https://godoc.org/github.com/golang/protobuf/ptypes.

`Timestamp` maps to `time.Time` fairly well, and `Duration` to `time.Duration` too. Let's support them.

A bit tricky, since gunk files can't import standard library packages right now.

mvdan:
Mapping to their google/protobuf counterparts, which can be converted
back and forth with the github.com/golang/protobuf/ptypes Go package.

Using these standard Go types feels more Go-like (or Gunk-like) than
using custom names for this, or even importing the google proto files
directly.

Fixes #26.


mvdan:
I'm blocked on this to submit more PRs (because there's no chaining of them, like on Gerrit), so I'll merge this now. @vishen if you have any comments, we can address at a later time.

PR #39, opened 2018-11-01
loader: force position numbers to be on struct field tags

### LINKED ISSUE #34: Gunk should check that the protobuf field numbers are set in a struct
I forgot to add the `pb:"<field_number>`" to the gunk struct fields; protoc compiled, but when starting up the grpc client and server there was runtime errors (naturally). 

We should error if the `pb:"<field_number>`" is missing for now.

vishen:
Struct fields are now required to have a 'pb' tag on them. This tag
indicates the position number of the field to Protobuf.

Fixes #34


loader/util.go
```diff
@@ -43,12 +43,14 @@ func splitGunkTag(text string) (doc, tag string) {
 	return
 }
 
-func protoNumber(fieldTag *ast.BasicLit) *int32 {
-	if fieldTag == nil {
-		return nil
+func protoNumber(tag reflect.StructTag) (*int32, error) {
+	pbTag := tag.Get("pb")
+	if pbTag == "" {
+		return nil, fmt.Errorf("pb tag should not be empty")
```
mvdan: `pb field tag must be set`

loader/loader.go
```diff
@@ -377,20 +377,44 @@ func (l *Loader) convertMessage(tspec *ast.TypeSpec) (*desc.DescriptorProto, err
...
 			return nil, fmt.Errorf("unsupported field type: %v", ftype)
 		}
+
+		// Check that the struct field has a tag. We currently
+		// require all struct fields to have a tag; this is used
+		// to assign the position number for a field, ie: `pb:"1"`
+		if field.Tag == nil {
+			return nil, fmt.Errorf("missing required tag on %s", fieldName)
+		}
+		str, err := strconv.Unquote(field.Tag.Value)
```
mvdan: I was skipping the error here on purpose - we've parsed the file, so this can't error.

loader/loader.go
```diff
@@ -377,20 +377,44 @@ func (l *Loader) convertMessage(tspec *ast.TypeSpec) (*desc.DescriptorProto, err
...
+		// to set them to? Possibly pull out a json tag or something?
 		msg.Field = append(msg.Field, &desc.FieldDescriptorProto{
-			Name:     proto.String(field.Names[0].Name),
-			Number:   protoNumber(field.Tag),
+			Name:     proto.String(fieldName),
+			Number:   num,
 			TypeName: &tname,
 			Type:     &ptype,
 			Label:    &plabel,
+			JsonName: nil,
```
mvdan: remove?

loader/loader.go
vishen: Alright, I'll add a comment so this doesn't confuse me again.

loader/loader.go
vishen: Argh, that was in preparation for the json fields, I'll remove.

loader/util.go
vishen: Removed.

Issue #33, opened 2018-11-01
Gunk generated code currently different from Proto generated code


vishen:
Gunk currently is missing the `opt` option in the `protobuf` tag, this is because we don't return an optional label when converting go types. 

The `protobuf` tag for "name" is different as well, not sure if this is an issue and so far it hasn't caused any issues.

The `json` tag is using whatever the field name is, however this isn't idiomatic json naming; we should convert the name to camel_case if there is no json tag on the gunk struct.

Gunk generated code:
```
type Result struct {
	Success              bool     `protobuf:"varint,1,,name=Success,proto3" json:"Success,omitempty"`
	ErrorCode            string   `protobuf:"bytes,2,,name=ErrorCode,proto3" json:"ErrorCode,omitempty"`
	ErrorMessage         string   `protobuf:"bytes,3,,name=ErrorMessage,proto3" json:"ErrorMessage,omitempty"`
	XXX_NoUnkeyedLiteral struct{} `json:"-"`
	XXX_unrecognized     []byte   `json:"-"`
	XXX_sizecache        int32    `json:"-"`
}
```

Proto generated code:
```
type Result struct {
	Success              bool     `protobuf:"varint,1,opt,name=success,proto3" json:"success,omitempty"`
	ErrorCode            string   `protobuf:"bytes,2,opt,name=error_code,json=errorCode,proto3" json:"error_code,omitempty"`
	ErrorMessage         string   `protobuf:"bytes,3,opt,name=error_message,json=errorMessage,proto3" json:"error_message,omitempty"`
	XXX_NoUnkeyedLiteral struct{} `json:"-"`
	XXX_unrecognized     []byte   `json:"-"`
	XXX_sizecache        int32    `json:"-"`
}
```

vishen:
The options issue was fixed with https://github.com/gunk/gunk/pull/40

vishen:
There is still some differences, but they are expected differences now:

Gunk:
```
type Message struct {
	Msg                  string   `protobuf:"bytes,1,opt,name=Msg,json=msg,proto3" json:"Msg,omitempty"`
	Code                 int32    `protobuf:"varint,2,opt,name=Code,json=code,proto3" json:"Code,omitempty"`
	XXX_NoUnkeyedLiteral struct{} `json:"-"`
	XXX_unrecognized     []byte   `json:"-"`
	XXX_sizecache        int32    `json:"-"`
}
```

Proto:
```
type Message struct {
[...]

PR #41, opened 2018-11-02
loader: set json name for struct field

vishen:
If a struct field has a 'json' tag, we should use that to set the
'JsonName' on the proto field descriptor. This means that when using
'jsonpb' it will marshal and unmarshal using this name.

Updates #33


mvdan:
I think yours is more readable and more Go-like, especially since that's how one would write Go code normally.

The only possible gripe I have with this is that the json tag isn't kept as-is, i.e. using encoding/json on the output won't work. But that's fine, and the same happens with the pb tag.

PR #43, opened 2018-11-02
loader: change the proto package to be the gunk package

vishen:
Change the proto file descriptor package to be the same as the Gunk
package. Previously the proto file descriptor package was the path of
the Gunk file, it will now be whatever is defined as the Gunk package.


loader/loader.go
```diff
@@ -632,6 +633,19 @@ func (l *Loader) convertEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, er
 	return enum, nil
 }
 
+// formatTypeForPackage will format the type name for that package. If the
+// package is nil, it will format the type for the current file / package that is
+// being processed.
+//
+// Currently we format the type as ".<pkg_name>.<type_name>"
+func (l *Loader) formatTypeForPackage(typeName string, pkg *types.Package) string {
```
mvdan: I think that a name like `qualifiedTypeName` would be more descriptive of what the purpose here is.

loader/loader.go
```diff
@@ -632,6 +633,19 @@ func (l *Loader) convertEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto, er
 	return enum, nil
 }
 
+// formatTypeForPackage will format the type name for that package. If the
+// package is nil, it will format the type for the current file / package that is
```
mvdan: I'd drop the `file / package` bit in favor of just either option, as that makes the sentence unnecessarily complex to parse. We care about the package name, so I'd just say "current package".

loader/loader.go
vishen: Done.

loader/loader.go
vishen: Done.

PR #53, opened 2018-11-07
convert: started proto to gunk converter

vishen:
Add a convert command that will convert proto files to gunk files. The
convert command takes in a single .proto file, and converts it to a
.gunk file.

The converter will convert most proto specific declarations; enums,
messages, services, package and package options. Most message field
types should be converted as well, including the basic (scalar) proto
types and maps and arrays.

Any nested enums in messages are converted to top level declarations
because Gunk doesn't currently support nested structures.

The google http annotation is converted to +gunk equivalant.

We use github.com/emicklei/proto to parse the proto file, rather than
using protoc, because protoc doesn't keep the comments when generating a
proto filedescriptor.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -1,0 +1,430 @@
+package convert
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/scanner"
+
+	"github.com/emicklei/proto"
+	"github.com/knq/snaker"
+)
+
+// pkgOpt stores a package option and its comments.
+// eg: "go_package" or "java_package".
+type pkgOpt struct {
+	typeName string // go_package or java_package, etc.
+	value    string
+	comment  *proto.Comment
+}
+
+type builder struct {
+	// The converted proto to gunk declarations. This only stores
+	// the messages, enums and services. These get converted to as
+	// they are found.
+	translatedDeclarations []string
+
+	// The package, option and imports from the proto file.
+	// These are converted to gunk after we have converted
+	// the rest of the proto declarations.
+	pkg     *proto.Package
+	pkgOpts []pkgOpt
+	imports []*proto.Import
+}
+
+// Convert will convert a proto file to a gunk file,
+// saving the file in the same folder as the proto file.
+func Convert(path string, overwrite bool) error {
+	if filepath.Ext(path) != ".proto" {
+		return fmt.Errorf("convert requires a .proto file")
+	}
+	reader, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("unable to read file %q: %v", path, err)
+	}
+	defer reader.Close()
+
+	// Parse the proto file.
+	parser := proto.NewParser(reader)
+	d, err := parser.Parse()
+	if err != nil {
+		return fmt.Errorf("unable to parse proto file %q: %v", path, err)
+	}
+
+	// Start converting the proto declarations to gunk.
+	b := builder{}
+	for _, e := range d.Elements {
+		if err := b.handleProtoType(e); err != nil {
+			return fmt.Errorf("%v\n", err)
+		}
+	}
+
+	// Convert the proto package and imports to gunk.
+	translatedPkg := b.handlePackage()
+	translatedImports := b.handleImports()
+
+	// Add the converted package and imports, and then
+	// add all the rest of the converted types. This will
+	// keep the order that things were declared.
+	w := &strings.Builder{}
+	w.WriteString(translatedPkg)
+	w.WriteString("\n\n")
+	w.WriteString(translatedImports)
+	w.WriteString("\n")
+	for _, l := range b.translatedDeclarations {
+		w.WriteString("\n")
+		w.WriteString(l)
+		w.WriteString("\n")
+	}
+
+	// TODO: We should run this through the Gunk generator to
+	// make sure that it compiles?
+	// TODO: Run this through the gunk fmt when it is implemented.
+
+	fileToWrite := strings.Replace(filepath.Base(path), ".proto", ".gunk", 1)
+	fullpath := filepath.Dir(path) + "/" + fileToWrite
+
+	if _, err := os.Stat(fullpath); !os.IsNotExist(err) && !overwrite {
+		return fmt.Errorf("path already exists %q, use --overwrite", fullpath)
+	}
+	if err := ioutil.WriteFile(fullpath, []byte(w.String()), 0644); err != nil {
+		return fmt.Errorf("unable to write to file %q: %v", fullpath, err)
+	}
+
+	return nil
+}
+
+// format will write output to a string builder, adding in indentation
+// where required. It will write the comment first if there is one,
+// and then write the rest.
+//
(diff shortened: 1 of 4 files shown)


converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+		fmt.Fprintf(w, "\t")
+	}
+	fmt.Fprintf(w, s, args...)
+}
+
+// emitError will save the current error position and error. The error
+// formatting will use these when reporting errors.
+func (b *builder) emitError(pos scanner.Position, s string, args ...interface{}) {
+	b.errPos = pos
+	b.err = fmt.Errorf(s, args...)
```
mvdan: what will happen if we encounter many errors? you should either only save the first error, or ensure that the entire program stops when encountering the first error. You could accomplish the second option with recovered panics, for example.

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+	typeName string // go_package or java_package, etc.
+	value    string
+	comment  *proto.Comment
+}
+
+type builder struct {
+	// The converted proto to gunk declarations. This only stores
+	// the messages, enums and services. These get converted to as
+	// they are found.
+	translatedDeclarations []*strings.Builder
```
mvdan: Seems oddly specific to store `*strings.Builder` here. Why not `string` or `fmt.Stringer`?

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+
+	// TODO: We should run this through the Gunk generator to
+	// make sure that it compiles?
+	// TODO: Run this through the gunk fmt when it is implemented.
+
+	fileToWrite := strings.Replace(filepath.Base(path), ".proto", ".gunk", 1)
+	fullpath := filepath.Dir(path) + "/" + fileToWrite
+
+	if _, err := os.Stat(fullpath); !os.IsNotExist(err) && !overwrite {
+		return fmt.Errorf("path already exists %q, cannot overwrite", fullpath)
```
mvdan: I'd say `use --overwrite` rather than `cannot overwrite`. It's a matter of choice, not ability :)

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+		return fmt.Errorf("unable to write to file %q: %v", fullpath, err)
+	}
+
+	return nil
+}
+
+// Write will write output to a string builder, adding in indentation
+// where required. It will write the comment first if there is one,
+// and then write the rest.
+func (b *builder) Write(w *strings.Builder, indent int, comments *proto.Comment, s string, args ...interface{}) {
```
mvdan: nit: don't call this `Write`, as it's easily confused with the method from `io.Writer`. You could call it `write`, `output`, etc.

Also, I'd handle indentation via an integer field. Then, you'd do `b.indent++` or `b.indent--` in certain places. That should be clearer and less repetitive.

mvdan: Also, `s` is a format string, so I'd call it `format` like `fmt.Printf`'s arguments.

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+	if repeated {
+		typ = "[]" + typ
+	}
+
+	// TODO(vishen): Is this correct to explicitly camelcase the variable name and
+	// snakecase the json name???
+	// If we do, gunk should probably have an option to set the variable name
+	// in the proto to something else? That way we can use best practises for
+	// each language???
+	b.Write(w, 1, comment, "%s %s `pb:\"%d\" json:\"%s\"`\n", snaker.ForceCamelIdentifier(name), typ, sequence, snaker.CamelToSnake(name))
```
mvdan: you've made the write method fairly low level, so I'd split these very long lines into multiple, like:

```
b.Write(w, 1, comment, "%s %s", snaker.ForceCamelIdentifier(name), typ)
b.Write(w, 1, comment, " `pb:\"%d\" json:\"%s\"`\n",  sequence, snaker.CamelToSnake(name))
```

converter/converter.go
```diff
@@ -0,0 +1,437 @@
+package converter
```
mvdan: nit: we've named the others `generate` and `format`, so I'd call this `convert`.

converter/converter.go
mvdan: (same applies to the types passed around by functions)

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+
+	return nil
+}
+
+// Write will write output to a string builder, adding in indentation
+// where required. It will write the comment first if there is one,
+// and then write the rest.
+func (b *builder) Write(w *strings.Builder, indent int, comments *proto.Comment, s string, args ...interface{}) {
+	if comments != nil {
+		for _, c := range comments.Lines {
```
mvdan: does this deal with inline comments, e.g.

```
some proto line // inline comment
```

I presume not, in which case we need to either error in those cases, or add a TODO to fix them

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+	return nil
+}
+
+// handleMessageField will convert a messages field to gunk.
+func (b *builder) handleMessageField(w *strings.Builder, field proto.Visitee) error {
+	var name string
+	var typ string
+	var sequence int
+	var repeated bool
+	var comment *proto.Comment
```
mvdan: these are quite a few vars, so I'd group them with `var ()` to align the types nicely.

converter/converter.go
vishen: It will bail out with the first error it finds. I didn't think this needed to handle more than one error, at least not to begin with anyway?

converter/converter.go
vishen: I had the indent initially as you described, but it was actually really confusing when viewing the code because you had to keep track of the current indent level or constantly try and find where it is indenting.

converter/converter.go
vishen: Ah, missed that, good catch! I have left a TODO. It might be a bit "hacky" to do, so I will deal with it later.

vishen:
I have addressed most of your comments @mvdan

converter/converter.go
```diff
@@ -0,0 +1,437 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"text/scanner"
+
+	"github.com/emicklei/proto"
+	"github.com/knq/snaker"
```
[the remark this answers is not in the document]
vishen: I remember seeing it a while ago, and thought it would be useful to use in here :)

converter/converter.go
mvdan: That's my point - it won't always bail when finding the first error. For example, `handleProtoType` is a function that can emit an error, return a nil error, and the program will continue converting.

The simplest fix would be for emitError to do nothing if b.err is already non-nil. I also find it a bit confusing how some functions can both return an error and emit an error.

mvdan:
Still find the error stuff kinda confusing, but otherwise LGTM - feel free to merge when ready

converter/converter.go
vishen: It should bail when it finds the first error, but I do see your point. Actually, the only purpose of this was to allow for better error messages (with the position). I will change it to return a formatted error rather than this emit error thing, and propagate the formatted error.

PR #57, opened 2018-11-12
loader: accurate position info for gunk tag errors

### LINKED ISSUE #35: Parsing '+gunk' syntax gives slightly unhelpful error message when it fails.
When the parsing of `+gunk` syntax fails, it would be helpful if it gave a better error message (if possible). For the below invalid gunk, it gives `error: eval:2:3: expected 'EOF', found Method` - it is useful if I knew it was parsing the below, but I had no idea. 

If we could point to the actual line number in the file or the function it is parsing, that might be enough to help people track down the issue.

```
        // +gunk http.Match
        //              Method: "POST",
        //              Path:   "/v1/echo",
        //              Body:   "*",
        //      }
        Echo(Message) Message
```

mvdan:
For now, do syntax errors only. Left a TODO to update the positions in
the returned ast.Expr, which will then fix type-checking errors. Leaving
that for a later time.

Fixes #35.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/parser"
+	"go/scanner"
 	"go/token"
 	"html/template"
 	"os"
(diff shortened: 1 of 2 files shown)


mvdan:
@vishen this is a fairly subtle fix - took me a few tries to get it right - so feel free to point out any confusing bits.

PR #58, opened 2018-11-13
config: added basic gunkconfig

vishen:
Added a config package to find and read a .gunkconfig. It behaves in a
similar way to go mod, which starts in the working directory and then
moves up each parent checking for a .gunkconfig.

The .gunkconfig can contain one or more 'generate' sections. Each
generate section is a single command to be run. Currently, the command
is assumed to be a protoc-gen-* command, which takes in the proto
filedescriptor as stdin and outputs a proto filedescriptor as stdout.


main.go
```diff
@@ -30,10 +31,15 @@ func main() {
 }
 
 func main1() int {
-	var err error
+	gunkConfig, err := config.Load("")
```
vishen: Should the config store the dir (which is `""` in this case)? If so maybe it should also store the patterns?

README.md
```diff
@@ -31,7 +31,8 @@ the Go programming language:
 Working with the `gunk` command line tool should be instantly recognizable to
 experienced Go developers:
 
-	gunk generate ./examples/util
+    cd ./examples/util
+	gunk generate .
```
mvdan: don't mix tabs and spaces, and also `gunk generate` can be used instead of `gunk generate .`

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/knq/ini"
+	"github.com/knq/ini/parser"
+)
+
+type KV struct {
```
mvdan: seems like a very short name for a global type - `KeyValue` instead?

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+	Params  []KV
+}
+
+type Config struct {
+	Dir        string
+	Patterns   []string
+	Generators []Generator
+}
+
+func Load(path string) (*Config, error) {
```
mvdan: I'd rename `path` to `wd` or `startdir` for clarity, especially because I initially read it as "config path"

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+				return nil, fmt.Errorf("could not find a .gunkconfig")
+			}
+			continue
+		}
+		defer reader.Close()
+		cfg, err = load(reader)
+		if err != nil {
+			return nil, fmt.Errorf("error loading %q: %v", configPath, err)
+		}
+		if cfg != nil {
```
mvdan: why would `load` ever return nil error and nil cfg? this case could also lead to weird behavior, such as parsing multiple config files and leaving them open (see the `defer`).

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	config := &Config{
+		Generators: make([]Generator, 0, len(f.AllSections())),
+	}
+	for _, s := range f.AllSections() {
+		var err error
+		var gen *Generator
+		switch s.Name() {
+		case "":
+			// This seems to happen sometimes?
```
mvdan: if this is a TODO, please clearly mark it as such

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil, fmt.Errorf("unknown section %q", s.Name())
+		}
+		if err != nil {
+			return nil, err
+		}
+		if gen != nil {
+			config.Generators = append(config.Generators, *gen)
+		}
+
+		gen = nil
```
mvdan: I'm pretty sure this line is unnecessary, since you declare the pointer variable at every iteration

config/config.go
```diff
@@ -0,0 +1,110 @@
...
+			// This seems to happen sometimes?
+		case "generate":
+			gen, err = handleGenerate(s)
+		default:
+			return nil, fmt.Errorf("unknown section %q", s.Name())
+		}
+		if err != nil {
+			return nil, err
+		}
+		if gen != nil {
```
mvdan: instead of this, I'd do a `continue` in the `TODO` case above. Then, there's no possibility for the weird `err == nil && gen == nil` case.

generate/generate.go
```diff
@@ -117,13 +121,52 @@ type Generator struct {
 // It is fine to pass the plugin.CodeGeneratorRequest to every protoc generator
 // unaltered; this is what protoc does when calling out to the generators and
 // the generators should already handle the case where they have nothing to do.
-func (g *Generator) GeneratePkg(path string) error {
+func (g *Generator) GeneratePkg(path string, gens []config.Generator) error {
 	req := g.requestForPkg(path)
+	// Run generator for go
```
mvdan: I imagined that `go` would be listed too, even if it's almost always going to be on, or even if we end up statically linking with it.

Is there a reason not to? If you left this for a later PR, just leave a TODO for clarity.

generate/generate.go
```diff
@@ -59,7 +61,9 @@ func Generate(dir string, patterns ...string) error {
 	for _, pkg := range pkgs {
 		g.gunkPkgs[pkg.PkgPath] = pkg
 		for i, name := range pkg.GunkNames {
-			g.origPaths[name] = pkg.GunkFiles[i]
+			// Save only the directory for each file path.
+			dir, _ := filepath.Split(name)
+			g.origPaths[dir], _ = filepath.Split(pkg.GunkFiles[i])
```
mvdan: I'm not sure why this change is necessary - you seem to do the same `filepath.Split` when querying and using `origPaths`, so it seems to me like the old way would be simpler

main.go
mvdan: I think the config should keep the path it was found at - e.g. `/home/mike/src/brankas-proto/.gunkconfig`. It can be useful to report errors at a later time, such as `found unknown generator "foo" in .../foo/bar/.gunkconfig`.

I don't think keeping the search start directory as part of the config struct will be helpful, unless you have a specific use case for it.

main.go
```diff
@@ -30,10 +31,15 @@ func main() {
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		return 1
+	}
 	switch kingpin.MustParse(app.Parse(os.Args[1:])) {
 	case gen.FullCommand():
-		err = generate.Generate("", *genPatterns...)
+		gunkConfig.Patterns = *genPatterns
+		err = generate.Generate(gunkConfig)
```
mvdan: I was actually imagining that each subcommand would parse the config separately. Reasons:

* Not all subcommands will need the config (think `help`, `version`, likely `gunk fmt`, etc)
* The subcommand interface should revolve around a list of arguments (strings) directly from the command line, leaving it up to the subcommand to parse flags and interpret the parameters
* Should only be four lines of simple code for each of the subcommands that need a config

I realise that the old `Generate` uses a workdir parameter instead of `os.Getwd`. That was simply so that I wouldn't have to rely on a hacky `os.Chdir` in the tests. Perhaps there's a better way to deal with that.

README.md
vishen: Opps, I'll fix that!

generate/generate.go
vishen: The old way includes the filename, which makes it kind of hard to match when getting the modifed output filenames from a protoc-gen-* generator. Since this is only used to know where the package path points to on disk, it is easier to just check the "directory" path since that won't change between `protoc-gen-*` runs, rather than changing `github.com/gunk/gunk/examples/utils/echo.pb.gw.go` to `github.com/gunk/gunk/examples/utils/echo.gunk` and then doing the lookup for the on-disk location.

generate/generate.go
vishen: This is currently an issue with the ini parser, it doesn't handle multiple sections with the same name and getting the key values from the sections correctly. I need to submit a PR to handle the multiple sections to the ini package. However, this should likely be changed to use the library rather than shelling out, I'll just add that as a TODO.

main.go
vishen: Yeah, I jumped between the two. A couple of approaches I had the config being parsed from each subcommand, but I ended up going with this because I could quite see a situation where the config file parsing would be different between the subcommands. However, what you say makes sense, I'll make the change.

generate/generate.go
mvdan: Ah gotcha - you're probably right.

What we should do in the long run is use the dir where the gunk file is as the default output path, and allowing the gunk config to override that. For example, in brankas-proto, `pb/v2/*` is placed in `v2/*`.

README.md
vishen: I tried running with `gunk generate` and I get the error `error: no Gunk packages to generate`.

mvdan: Huh... please leave it as it is then, and file an issue :)

vishen:
@mvdan PTAL, I think I handled all the comments? :)

README.md
vishen: Done!

PR #61, opened 2018-11-13
all: remove stutter from subcommand funcs

mvdan:
generate.Run is better than generate.Generate.


mvdan:
The hard link PR is up separately :) then I'll rebase this.

PR #63, opened 2018-11-13
config,generate: use protoc to generate other language proto bindings

### LINKED ISSUE #46: Generate protobuf files for multiple languages
Currently Gunk only generates protobuf code for Go, but we should add support for other languages.  [Protoc will generate most of the languages for us](https://github.com/protocolbuffers/protobuf#protobuf-runtime-installation) without any plugins:

- C++
- Java
- Python
- Javascript
- Ruby
- Others...

We can piggyback on `protoc` to generate the languages and look at an alternative to `protoc` down the track.  

`protoc` has an command line options for taking in `FileDescriptorSet`'s and using those:

```
 --descriptor_set_in=FILES   Specifies a delimited list of FILES
                              each containing a FileDescriptorSet (a
                              protocol buffer defined in descriptor.proto).
                              The FileDescriptor for each of the PROTO_FILES
                              provided will be loaded from these
                              FileDescriptorSets. If a FileDescriptor
                              appears multiple times, the first occurrence
                              will be used.
```

and we just need to make sure we pass through the command line arguments for the languages we want to generate: `--js_out`, `--python_out`, etc.

vishen:
Use protoc to generate the other language bindings for now. We can use
protoc to generate most other languages, and for those protoc can't
generate, we will need to use protoc-gen-* plugins.

This PR adds a new config paramater to gunkconfg, 'protoc=<language>'.

'protoc' is more strict than 'protoc-gen-go'; If an enum is found without
any values, we need ignore the enum. If a typeName to a message field is
empty, we need to set the typeName to 'nil', rather than a pointer to
an empty string.

Fixes #46


generate/generate.go
```diff
@@ -657,6 +733,10 @@ func (g *Generator) convertEnum(tspec *ast.TypeSpec) (*desc.EnumDescriptorProto,
 		}
 	}
 	g.enumIndex++
+	// If an enum doesn't have any values
+	if len(enum.Value) == 0 {
```
vishen: This is required because protoc is more strict that protoc-gen-go. protoc won't compile the proto if there is "hanging" enums. We should either error out if there is no enum values, or ignore that enum. This is currently happening due to the `type`s in https://github.com/gunk/opt/blob/master/opt.gunk

generate/generate.go
```diff
@@ -442,7 +518,7 @@ func (g *Generator) convertMessage(tspec *ast.TypeSpec) (*desc.DescriptorProto,
 		msg.Field = append(msg.Field, &desc.FieldDescriptorProto{
 			Name:     proto.String(fieldName),
 			Number:   num,
-			TypeName: &tname,
+			TypeName: protoStringOrNil(tname),
```
vishen: protoc was failing because the type_name was `""` rather than not being set.

generate/generate.go
```diff
@@ -134,19 +134,91 @@ func (g *Generator) GeneratePkg(path string, gens []config.Generator) error {
...
-	req.Parameter = proto.String(strings.Join(protoParameters, ","))
+	if err := tmpfile.Close(); err != nil {
+		return fmt.Errorf("unable to close tmp file: %v", err)
+	}
+
+	// Build up the protoc command line arguments.
+	command := "protoc"
+	args := []string{
+		fmt.Sprintf("--%s_out=%s", gen.ProtocGen, gen.ParamsWithOut()),
+		fmt.Sprintf("--descriptor_set_in=%s", protoFileDescriptorPath),
```
mvdan: I'd use `/dev/stdin` instead. It's not portable to Windows, but we don't care about that for now. Also, we use `/dev/stdout` elsewhere :)

I did report thiss to upstream ages ago, but unfortunately it has had zero attention so far: https://github.com/protocolbuffers/protobuf/issues/4163

generate/generate.go
```diff
@@ -362,11 +434,16 @@ func (g *Generator) translateDecl(decl ast.Decl) error {
 			}
 			g.pfile.Service = append(g.pfile.Service, srv)
 		case *ast.Ident:
+			// TODO(vishen): Check to see if the ident is a known file
+			// option and add if to  `p.file.Options`.
```
mvdan: please no backticks in comments - they're not markdown.

generate/generate.go
mvdan: SGTM; this was an oversight on my part anyway.

generate/generate.go
```diff
@@ -134,19 +134,91 @@ func (g *Generator) GeneratePkg(path string, gens []config.Generator) error {
...
+	// package path, but it would be created from the current relative
+	// directory, rather than an absolute path.
+	//
+	// We also replace .gunk with .proto. protoc will turn
+	// 'echo.gunk' into 'echo.gunk_pb.js' which makes it a bit
+	// hard to replace afterwards. This seemed like the easier approach.
+	for i, f := range fds.File {
+		for _, ftg := range req.GetFileToGenerate() {
+			if f.GetName() == ftg {
+				_, name := filepath.Split(ftg)
```
mvdan: Simpler: `name := filename.Base(ftg)`

generate/generate.go
```diff
@@ -134,19 +134,91 @@ func (g *Generator) GeneratePkg(path string, gens []config.Generator) error {
...
+	// to determine the filename. In our case, that would be the
+	// package path, but it would be created from the current relative
+	// directory, rather than an absolute path.
+	//
+	// We also replace .gunk with .proto. protoc will turn
+	// 'echo.gunk' into 'echo.gunk_pb.js' which makes it a bit
+	// hard to replace afterwards. This seemed like the easier approach.
+	for i, f := range fds.File {
+		for _, ftg := range req.GetFileToGenerate() {
+			if f.GetName() == ftg {
```
mvdan: invert and early continue to unindent the rest of the loop body

generate/generate.go
vishen: Is there a benefit to using stdin here? We create and delete temp files when using protoc to load the imports, and as you said, it may make it _possibly_ windows compatible.

mvdan: Both `/dev/` and temporary files are hacks - I'm simply saying we should be consistent. Elsewhere in the code we already use `--descriptor_set_out=/dev/stdout`, so I think you should do the equivalent here.

Unless you prefer to use temporary files in both. We don't even test on Windows, so I personally think it's not a priority to guesstimate what will work on Windows.

generate/generate.go
vishen: Fixed.

generate/generate.go

generate/generate.go

generate/generate.go
vishen: Yup, makes sense. I have updated it to use /dev/stdin.

PR #64, opened 2018-11-14
generate, mod: copy pointer values before updating them

vishen:
In each 'generateProtoc' run we modify the protofile name, however, this
was a pointer, and was changing the value for all runs of
'generateProtoc' and 'generatePlugin'. Make a dumb copy of the protofile
before we use it.

Updated the ini parser to a forked version that handles multiple
section names, and getting keys from the different sections, correctly.

I have created a PR upstream for [ini](https://github.com/knq/ini/pull/1), until then we can use my fork of the ini parser.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -152,8 +152,17 @@ func (g *Generator) generateProtoc(req plugin.CodeGeneratorRequest, gen config.G
 	// a formatted list of what is in req.FileToGenerate.
 	protoFilenames := []string{}
 
+	// Make copies of the req.ProtoFile so we don't accidentally change
+	// the values of the pointers which will affect all other protoc
+	// generator runs.
+	protoFiles := make([]*desc.FileDescriptorProto, len(req.ProtoFile))
+	for i, pf := range req.ProtoFile {
+		tmpPf := *pf
+		protoFiles[i] = &tmpPf
+	}
+
 	fds := desc.FileDescriptorSet{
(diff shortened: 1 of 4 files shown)


generate/generate.go
```diff
@@ -152,8 +152,17 @@ func (g *Generator) generateProtoc(req plugin.CodeGeneratorRequest, gen config.G
 	// a formatted list of what is in req.FileToGenerate.
 	protoFilenames := []string{}
 
+	// Make copies of the req.ProtoFile so we don't accidentally change
+	// the values of the pointers which will affect all other protoc
+	// generator runs.
+	protoFiles := make([]*desc.FileDescriptorProto, len(req.ProtoFile))
```
vishen: I am sure there is a better way to do this..?

mvdan: If you only wanted to copy the array, you could do `list2 := append([]T{}, list1...)`. However, you want to copy each pointer's backing data too, so I don't think there's a short-hand for this.

Perhaps we could make the slice a `[]desc.FileDescriptorProto` instead. But I don't think that's worth it, just to avoid three lines of code.

vishen:
I have updated the forked ini package to the latest one that is passing all tests.

mvdan:
@vishen we can merge this btw; there's no reason why the upstream PR should block this one, as long as we remember to get rid of the replace directive later.

PR #65, opened 2018-11-14
config: use params key in section for generator params

vishen:
Have a specific key for generator params in gunkconfig. This is required
because some generators want the params in arbitrary formats, and the ini
section format of key values won't make sense for all generators.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,21 +5,15 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/knq/ini"
 	"github.com/knq/ini/parser"
 )
 
-type KeyValue struct {
-	Key   string
-	Value string
-}
-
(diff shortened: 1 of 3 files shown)


mvdan:
Oh, this is a bummer. The big advantage of this was that we'd end up with nice flat lists, like:

```
param1=value1
param2=value2
param3=value3
...
```

instead of the longer and less readable:

```
params=param1=value1,param2=value2,param3=value3,...
```

Which generators do this? I thought this was the common protoc way of passing options.

vishen:
Yup, it is a bit annoying. The javascript one does for sure, haven't evaluated many others, but since we want to use javascript now we either have to use this approach, or, have something like:

```
param1=value1
param2=
param3=value3
```

and that would be converted to `param1=value1,param2,param3=value3`. Not sure if there are other alternatives?

PR #79, opened 2018-11-16
config: add a default config

vishen:
Added a 'DefaultConfig' in the config package. This will run
'protoc-gen-go' and 'protoc-gen-grpc-gateway' by default. The
'DefaultConfig' will be used if no '.gunkconfig' can be foung in the
current directory, or any parent directories.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,14 @@ import (
 	"github.com/knq/ini/parser"
 )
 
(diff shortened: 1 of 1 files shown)


vishen:
I don't really understand what you are proposing regarding the tests, but I want to start I will look at adding tests for the entire config package soon. These will likely be easier as separate PR's though.

PR #86, opened 2018-11-19
config,generate: look for gunkconfig in each gunk package

### LINKED ISSUE #82: Look for .gunkconfig separately for each Gunk package
And start looking for the config file from the package's directory, instead of `$PWD`.

vishen:
Look for a .gunkconfig in each gunk package, if one isn't found then
look in each parent to see if one exists. The path for generators to
output generated files will be the following:

    - Next to the *.gunk files, if no 'out' key specified (the default)
    - Wherever the 'out' key specifies. If the path is absolute, we use
    that path. If it isn't absolute, then we join the 'out' with the
    directory containing the .gunkconfig

Added 'examples/util/v1' which is where the files should be generated.

Closes #82

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,8 +48,12 @@ func (g Generator) ParamString() string {
 	return strings.Join(params, ",")
 }
 
-func (g Generator) ParamStringWithOut() string {
-	outPath := g.OutPath()
+// ParamStringWithOut will return the generator paramaters formatted
+// for protoc, including where protoc should output the generated files.
+// It will use 'packageDir' if no 'out' key was set in the config.
+func (g Generator) ParamStringWithOut(packageDir string) string {
+	// If no out path was specified, use the package directory.
+	outPath := g.OutPath(packageDir)
 	params := g.ParamString()
 	if params == "" {
 		return outPath
@@ -57,11 +61,16 @@ func (g Generator) ParamStringWithOut() string {
 	return params + ":" + outPath
 }
 
-func (g Generator) OutPath() string {
-	if g.Out != "" {
+// Determine an out path for a generator to write generated files to.
+// It will use 'packageDir' if no 'out' key was set in the config.
+func (g Generator) OutPath(packageDir string) string {
+	if g.Out == "" {
+		return packageDir
+	}
+	if filepath.IsAbs(g.Out) {
 		return g.Out
 	}
-	return g.ConfigDir
+	return filepath.Join(g.ConfigDir, g.Out)
 }
 
 type Config struct {
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -30,10 +30,6 @@ import (
 // Run generates the specified Gunk packages via protobuf generators, writing
 // the output files in the same directories.
 func Run(dir string, args ...string) error {
-	cfg, err := config.Load(dir)
-	if err != nil {
-		return fmt.Errorf("unable to load gunkconfig: %v", err)
-	}
 	g := &Generator{
 		dir:  dir,
 		fset: token.NewFileSet(),
@@ -79,6 +75,10 @@ func Run(dir string, args ...string) error {
 
 	// Finally, run the code generators.
 	for _, pkg := range pkgs {
+		cfg, err := config.Load(pkg.Dir)
+		if err != nil {
+			return fmt.Errorf("unable to load gunkconfig: %v", err)
+		}
 		if err := g.GeneratePkg(pkg.PkgPath, cfg.Generators); err != nil {
 			return err
 		}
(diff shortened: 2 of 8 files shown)


config/config.go
```diff
@@ -49,19 +49,29 @@ func (g Generator) ParamString() string {
 }
 
 func (g Generator) ParamStringWithOut() string {
+	// If no out path was specified, use the package directory.
 	outPath := g.OutPath()
+	if outPath == "" {
+		outPath = g.ConfigDir
+	}
```
mvdan: But the config dir isn't always the package dir :)

config/config.go
```diff
@@ -49,19 +49,29 @@ func (g Generator) ParamString() string {
...
+// Determine an out path for a generator to write generated files to.
+// An empty string is returned if no 'out' key was set.
 func (g Generator) OutPath() string {
-	if g.Out != "" {
+	if g.Out == "" {
+		return ""
+	}
+	if filepath.IsAbs(g.Out) {
 		return g.Out
+	} else {
```
mvdan: you can drop this else and unindent the line below

examples/util/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
mvdan: Surely we want to not gitignore the other existing files in examples too?

generate/generate.go
```diff
@@ -248,8 +248,8 @@ func (g *Generator) generatePlugin(req plugin.CodeGeneratorRequest, gen config.G
 		gpkg := g.gunkPkgs[pkgPath]
 		data := []byte(*rf.Content)
 		dir := gpkg.Dir
-		if gen.Out != "" {
-			dir = gen.Out
+		if op := gen.OutPath(); op != "" {
```
mvdan: `op` generally means `operator`, so I'd change this variable to `path` or `outPath`.

testdata/scripts/config_default.txt
```diff
@@ -0,0 +1,35 @@
...
+    "github.com/gunk/opt/http"
+)
+
+type Message struct {
+	Msg string `pb:"1"`
+}
+
+type Util interface {
+	// Echo echoes a message.
+    //
```
mvdan: mixing tabs and spaces? :) I'd remove the http stuff from these config scripts, it's irrelevant and quite verbose.

vishen: Arghh, I copied and didn't replace the tabs. The http stuff is required however, as otherwise the `protoc-gen-grpc-gateway` won't produce any output files. And since that is part of the default config, I think we want to test that it works :)

config/config.go
vishen: Hmm, I guess we will have to pass in the gunk package path, and use that rather than `g.ConfigDir`

mvdan: > I guess we will have to pass in the gunk package path

assuming you mean the directory, not the import path, it sounds like an option :)

examples/util/.gitignore
vishen: Opps, this was supposed to be in `examples/util/v1`

config/config.go
vishen: Yes I am :) Done.

config/config.go
vishen: Done.

vishen:
I have added the test for a gunkconfig in a parent directory, and hopefully addressed all the comments.

config/config.go
```diff
@@ -48,20 +48,29 @@ func (g Generator) ParamString() string {
 	return strings.Join(params, ",")
 }
 
-func (g Generator) ParamStringWithOut() string {
-	outPath := g.OutPath()
+// ParamStringWithOut will return the generator paramaters formatted
+// for protoc, including where protoc should output the generated files.
+// It will use 'gunkPackage' if no 'out' key was set in the config.
+func (g Generator) ParamStringWithOut(gunkPackage string) string {
```
mvdan: assuming the parameter here is the package's directory, I'd call this `packageDir`

examples/util/v1
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
mvdan: now this is just a file called `v1`?

config/config.go
vishen: Done.

generate/generate.go
```diff
@@ -203,12 +212,14 @@ func (g *Generator) generateProtoc(req plugin.CodeGeneratorRequest, gen config.G
...
 	command := "protoc"
 	args := []string{
-		fmt.Sprintf("--%s_out=%s", gen.ProtocGen, gen.ParamStringWithOut()),
+		fmt.Sprintf("--%s_out=%s", gen.ProtocGen, gen.ParamStringWithOut(protocOutputPath)),
 		"--descriptor_set_in=/dev/stdin",
 	}
 
 	args = append(args, protoFilenames...)
 
+	fmt.Println(args)
```
mvdan: :)

vishen: Removed :)

PR #87, opened 2018-11-19
config: merge gunkconfigs

### LINKED ISSUE #83: Support appending multiple Gunk configs found
For example, when running `gunk generate ./foo/bar`, if `./foo/bar/.gunkconfig` and `./foo/.gunkconfig` exist, the generate sections should be appended to form the package's generate config.

vishen:
Gather .gunkconfigs starting from the gunk package and looking in each
parent directory for .gunkconfigs to use.

Look for a .gunkconfig in parent directories until a 'go.mod' is found,
and we assume that to be the project root.

Closes #83


testdata/scripts/config_merge.txt
```diff
@@ -0,0 +1,37 @@
+env HOME=$WORK/home
+
+# Gunk generate with out specified in config
+gunk generate ./gunk
+
+exec ls
+exec ls gunk
```
mvdan: remove these two?

config/config.go
```diff
@@ -78,42 +81,78 @@ type Config struct {
 // working directory.
 func Load(dir string) (*Config, error) {
 	var err error
+	// Expand dir to the absolute path
```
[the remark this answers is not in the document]
vishen: I was initially having some problems with paths that had "." in them, but this isn't required anymore, I'll remove it.

testdata/scripts/config_merge.txt
vishen: Removed!

main_test.go
```diff
@@ -116,20 +116,8 @@ func generatedFiles(t *testing.T, dir string) (map[string]string, error) {
 	return files, err
 }
 
-func TestGenerateError(t *testing.T) {
+func TestGenerateScripts(t *testing.T) {
 	testscript.Run(t, testscript.Params{
 		Dir: filepath.Join("testdata", "scripts"),
-		Setup: func(env *testscript.Env) error {
```
vishen: This is no longer required because the .gunkconfig is being written as a file in the testscript, or it is using the default config.

main_test.go
```diff
@@ -116,20 +116,8 @@ func generatedFiles(t *testing.T, dir string) (map[string]string, error) {
 	return files, err
 }
 
-func TestGenerateError(t *testing.T) {
+func TestGenerateScripts(t *testing.T) {
```
vishen: Couldn't think of a better name, but `TestGenerateError` no longer seemed relevant as this seems to be testing both success and error cases?

PR #88, opened 2018-11-19
convert: run result through the formatter

### LINKED ISSUE #60: Use gunk fmt on the converted proto to gunk code
Since we now have gunk fmt, gunk fmt should be run on any converted proto to gunk code. Not sure how difficult this would be, but having nicely formatted gunk converted code would be good.

mvdan:
Add format.Source for this use case, mirroring the function signature in
go/format.

Fixes #60.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -2,6 +2,7 @@ package convert
 
 import (
 	"fmt"
+	"go/format"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


mvdan:
I couldn't think of an easy test for this. The current converter already produces well formatted code, and we haven't encountered a case where it did a bad job.

PR #95, opened 2018-11-20
generate: error on array function parameters

### LINKED ISSUE #81: Unable to return a list in a Gunk interface function
There seems to be a problem when trying to return a list from a gunk interface function. When generating the Go code for an interface function returning a list, it seems to return a single instance rather than a list. Haven't looked at it yet, so not sure what might be going on.

The following:
```
type DotService interface {
	//
	//
	// +gunk http.Match{
	//         Method: "GET",
	//         Path:   "/v1/dots",
	// }
	List() []Dot
}
```
produces:
```
type DotServiceServer interface {
	List(context.Context, *empty.Empty) (*Dot, error)
}
```

vishen:
Proto doesn't allow you to have repeated function parameters. However,
it is easy to forget when Go and other languages allow repeated function
parameters.

We now return an error if a functions input or output parameter is a
repeated structure.

Closes #81

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -644,9 +644,12 @@ func (g *Generator) convertParameter(tuple *types.Tuple) (*string, error) {
 		return nil, fmt.Errorf("multiple parameters are not supported")
 	}
 	param := tuple.At(0).Type()
-	_, _, tname := g.convertType(param)
+	_, label, tname := g.convertType(param)
 	if tname == "" {
 		return nil, fmt.Errorf("unsupported parameter type: %v", param)
+	}
+	if label == desc.FieldDescriptorProto_LABEL_REPEATED {
+		return nil, fmt.Errorf("parameter type should not be repeated")
 	}
 	return &tname, nil
 }
--- a/testdata/scripts/repeated_function_param.txt
+++ b/testdata/scripts/repeated_function_param.txt
@@ -1,0 +1,36 @@
+env HOME=$WORK/home
+
+! gunk generate ./p1
+stderr 'error: parameter type should not be repeated'
+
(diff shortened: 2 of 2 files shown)


generate/generate.go
```diff
@@ -644,10 +644,14 @@ func (g *Generator) convertParameter(tuple *types.Tuple) (*string, error) {
...
 	}
 	param := tuple.At(0).Type()
-	_, _, tname := g.convertType(param)
+	_, label, tname := g.convertType(param)
 	if tname == "" {
 		return nil, fmt.Errorf("unsupported parameter type: %v", param)
 	}
+	if label == desc.FieldDescriptorProto_LABEL_REPEATED {
+		return nil, fmt.Errorf("parameter type should not be repeated")
+
```
mvdan: remove empty line

generate/generate.go
```diff
@@ -644,10 +644,14 @@ func (g *Generator) convertParameter(tuple *types.Tuple) (*string, error) {
 		return nil, fmt.Errorf("multiple parameters are not supported")
 	}
 	param := tuple.At(0).Type()
-	_, _, tname := g.convertType(param)
+	_, label, tname := g.convertType(param)
```
mvdan: hmm, maybe we should have made it harder to skip part of the returned info...

generate/generate.go
vishen: Done!

generate/generate.go
vishen: Not too sure what we could do to make it harder to skip?

mvdan: Not sure either, hence the "hmm, maybe" :) just thinking outloud.

PR #97, opened 2018-11-20
all: added debug information controlled by command line flags

### LINKED ISSUE #91: Add -v -x command-line flags
Add `-v` and `-x` command-line flags, to provide better output (following Go's `-v` and `-x`) on generation:

```sh
$ gunk generate -v -x
```

Additionally, in my opinion if we add `-v` we can capture the `protoc` Warnings (see below) and only output them when `-v` is enabled.

```sh
~/src/go/src/github.com/gunk/gunk-example-server$ gunk generate ./...
all.proto: warning: Import github.com/gunk/opt/all.proto but not used.
all.proto: warning: Import github.com/gunk/opt/http/all.proto but not used.
all.proto: warning: Import github.com/gunk/opt/all.proto but not used.
all.proto: warning: Import github.com/gunk/opt/http/all.proto but not used.
all.proto: warning: Import github.com/gunk/opt/all.proto but not used.
all.proto: warning: Import github.com/gunk/opt/http/all.proto but not used.
```

vishen:
Added '-x' and '-v' to 'gunk', they behave similar to Go. '-x' will
print any commands that are run, these currently include 'protoc',
'protoc-gen-*' and 'ls' commands. '-v' will print when a gunk package
has been generated, this currently also doubles as a 'verbose' like
flag, which will print out any stderr from a command.

Closes #91

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -16,6 +16,8 @@ import (
 	"github.com/golang/protobuf/proto"
 	desc "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"golang.org/x/tools/go/packages"
+
+	"github.com/gunk/gunk/log"
 )
 
 // Load loads the Gunk packages on the provided patterns from the given dir and
@@ -174,7 +176,7 @@ syntax = "proto3";
 	defer os.Remove("gunk-proto")
 
 	// TODO: any way to specify stdout while being portable?
-	cmd := exec.Command("protoc", "-o/dev/stdout", "--include_imports", "gunk-proto")
+	cmd := log.ExecCommand("protoc", "-o/dev/stdout", "--include_imports", "gunk-proto")
 	out, err := cmd.Output()
 	if err != nil {
 		if e, ok := err.(*exec.ExitError); ok {
--- a/main.go
+++ b/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gunk/gunk/convert"
 	"github.com/gunk/gunk/format"
 	"github.com/gunk/gunk/generate"
+	"github.com/gunk/gunk/log"
 )
 
 var (
@@ -30,6 +31,9 @@ func main() {
 }
 
 func main1() int {
+	gen.Flag("print-commands", "print the commands").Short('x').BoolVar(&log.PrintCommands)
+	gen.Flag("verbose", "print the names of packages as they are generated").Short('v').BoolVar(&log.Verbose)
+
 	var err error
 	switch kingpin.MustParse(app.Parse(os.Args[1:])) {
 	case gen.FullCommand():
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -9,8 +9,6 @@ import (
 	"go/token"
 	"go/types"
 	"io/ioutil"
-	"os"
-	"os/exec"
 	"path/filepath"
 	"reflect"
 	"sort"
@@ -25,6 +23,7 @@ import (
 
 	"github.com/gunk/gunk/config"
 	"github.com/gunk/gunk/loader"
+	"github.com/gunk/gunk/log"
 )
 
 // Run generates the specified Gunk packages via protobuf generators, writing
@@ -82,6 +81,8 @@ func Run(dir string, args ...string) error {
 		if err := g.GeneratePkg(pkg.PkgPath, cfg.Generators); err != nil {
 			return err
 		}
+		log.PackageGenerated(pkg.PkgPath)
+
 	}
 	return nil
 }
@@ -218,9 +219,8 @@ func (g *Generator) generateProtoc(req plugin.CodeGeneratorRequest, gen config.G
 
 	args = append(args, protoFilenames...)
 
-	cmd := exec.Command(command, args...)
+	cmd := log.ExecCommand(command, args...)
 	cmd.Stdin = bytes.NewReader(bs)
-	cmd.Stderr = os.Stderr
 	out, err := cmd.Output()
 	if err != nil {
 		return fmt.Errorf("error executing %q: %q, %v", command, out, err)
@@ -235,9 +235,8 @@ func (g *Generator) generatePlugin(req plugin.CodeGeneratorRequest, gen config.G
 		return err
 	}
 
-	cmd := exec.Command(gen.Command)
+	cmd := log.ExecCommand(gen.Command)
 	cmd.Stdin = bytes.NewReader(bs)
-	cmd.Stderr = os.Stderr
 	out, err := cmd.Output()
 	if err != nil {
 		return fmt.Errorf("error executing %s: %s, %v", gen.Command, out, err)
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,8 @@ require (
 	github.com/rogpeppe/go-internal v1.0.0
 	github.com/stretchr/testify v1.2.2 // indirect
 	golang.org/x/net v0.0.0-20181029044818-c44066c5c816
-	golang.org/x/tools v0.0.0-20181030000716-a0a13e073c7b
+	golang.org/x/sys v0.0.0-20180905080454-ebe1bf3edb33 // indirect
+	golang.org/x/tools v0.0.0-20181120060634-fc4f04983f62
 	google.golang.org/genproto v0.0.0-20181029155118-b69ba1387ce2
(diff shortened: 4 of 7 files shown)


main.go
```diff
@@ -26,6 +29,11 @@ var (
 )
 
 func main() {
+	app.PreAction(func(_ *kingpin.ParseContext) error {
+		debug.SetPrintCommands(*debugCommands)
+		debug.SetPackageGenerated(*debugPackageGenerated)
+		return nil
+	})
 	os.Exit(main1())
 }
```
mvdan: `main` should only call `main1` - so the extra stuff should go in the other func

main.go
```diff
@@ -7,12 +7,15 @@ import (
...
+	"github.com/gunk/gunk/debug"
 	"github.com/gunk/gunk/format"
 	"github.com/gunk/gunk/generate"
 )
 
 var (
-	app = kingpin.New("gunk", "Gunk Unified N-terface Kompiler command-line tool.")
+	app                   = kingpin.New("gunk", "Gunk Unified N-terface Kompiler command-line tool.")
+	debugCommands         = app.Flag("print-commands", "print the commands").Short('x').Bool()
+	debugPackageGenerated = app.Flag("print-package-generated", "print the names of packages as they are generated").Short('v').Bool()
```
mvdan: I think these should be on a per-command basis, much like config loading. For example, why should `gunk format -x` make sense, given that it's never going to run any external program?

Similarly, we want `gunk generate -v -x`, not `gunk -v -x generate`. Particularly since it's very unusual for Go programas to take any arguments after flags.

loader/loader.go
```diff
@@ -173,15 +177,20 @@ syntax = "proto3";
 	}
 	defer os.Remove("gunk-proto")
 
+	debug.PrintCommand("protoc", "-o/dev/stdout", "--include_imports", "gunk-proto")
```
mvdan: Something that I don't like that much about this approach is how we need to:

* Remember to put a `debug.PrintCommand` next to every `exec.Command`
* Keep the two string lists in sync

How about adding a wrapper for `exec.Command`, which would:

* log the command to run, if necessary
* use `exec.Command`
* redirect its stderr to `os.Stdout`, if necessary

Then, if a user needs to modify the `exec.Cmd` such as setting `Stdin`, they can still do that.

As for the name, `debug.ExecCommand` is the best I can come up with. I think there's a better name than `debug` for this, since it's more than just debugging. Regular users could also benefit, so perhaps `log` would be a better package name.

debug/debug.go
```diff
@@ -0,0 +1,42 @@
...
+		return
+	}
+	fmt.Println(pkgPath)
+}
+
+func SetPrintCommands(value bool) {
+	printCommands = value
+}
+
+func SetPackageGenerated(value bool) {
```
mvdan: Any reason not to export the variables and remove the setters? We're the only users, so it's not like we need to protect anything.

loader/loader.go
mvdan: My suggestion above would keep everything working, with the caveat that the command's stderr would be streamed to the console, as opposed to buffered and then printed all at once. But that seems fine - better, even - to me.

mvdan:
Another thought - some pieces of the codebase use `cmd.Output()` and then `err.Stderr` - beware about those. You're touching `cmd.Stderr`, so I think `err.Stderr` will always be empty in those cases. Not sure what the best solution would be right now.

vishen:
> You're touching cmd.Stderr, so I think err.Stderr will always be empty in those cases

This is true. The error will still be printed to the console, just not as apart of the error message. We only lose stderr when we set the flag, and that flag is only set when redirecting the output to the terminal. This isn't great, but at least the stderr output isn't being lost completely.

debug/debug.go
vishen: Nope, just my personal preference. Changes to globals.

loader/loader.go
vishen: This should be done (if I understood what you were suggesting correctly)

main.go
vishen: Done.

main.go
vishen: Done.

loader/loader.go
```diff
@@ -83,6 +85,7 @@ func toGunkPackage(fset *token.FileSet, lpkg *packages.Package) (*GunkPackage, e
 		}
 	}
 
+	log.Command("ls", pkgDir, "*.gunk")
```
mvdan: This isn't really a command being run, so I wonder if we should log it at all. `go build -x` does print some commands such as when it creates a directory, but not when it loads Go files or lists directories.

I think the general rule to follow should be "can this action modify something or fail in dozens of ways"? I don't think `ls` falls under that.

main.go
```diff
@@ -9,13 +9,16 @@ import (
...
 
 var (
 	app = kingpin.New("gunk", "Gunk Unified N-terface Kompiler command-line tool.")
 
-	gen         = app.Command("generate", "Generate code from Gunk packages.")
-	genPatterns = gen.Arg("patterns", "patterns of Gunk packages").Strings()
+	gen                      = app.Command("generate", "Generate code from Gunk packages.")
+	genPatterns              = gen.Arg("patterns", "patterns of Gunk packages").Strings()
+	genDebugCommands         = gen.Flag("print-commands", "print the commands").Short('x').Bool()
+	genDebugPackageGenerated = gen.Flag("print-package-generated", "print the names of packages as they are generated").Short('v').Bool()
```
mvdan: Is there a way to not give long flag forms for these? I don't think anyone would use the long forms, and they're quite hard to remember anyway.

mvdan: Another option would be to leave `print-commands` alone and give the second one the long name `verbose`, which is more accurate.

main.go
```diff
@@ -33,6 +36,10 @@ func main1() int {
 	var err error
 	switch kingpin.MustParse(app.Parse(os.Args[1:])) {
 	case gen.FullCommand():
+		log.PrintCommands = *genDebugCommands
+		log.PrintPackageGenerated = *genDebugPackageGenerated
+		// Currently genDebugPackageGenerated doubles as the verbose flag
+		log.Verbose = *genDebugPackageGenerated
```
mvdan: If we were using `flag`, you could avoid this extra boilerplate via `flag.BoolVar(&log.PrintCommands, ...`. I'm sure that kingpin has something similar.

loader/loader.go
vishen: I thought I saw `go build -x` doing this, but I may have seen it wrong. I'll remove it.

main.go
vishen: Not that I could make work. Not having a long name gives errors. I'll change to verbose though.

main.go
vishen: I don't understand how what you are suggesting has less boilerplate than what is already there?

mvdan: Instead of:

```
someFlag = gen.Flag(...).Bool()
[...]
log.SomeFlag = *someFlag
```

You can just do

```
gen.Flag(...).BoolVar(&log.SomeFlag)
```

Of course you'd need to collapse `log.PrintPackageGenerated` into `log.Verbose`, which I think we should do anyway. For example, `go test -v` does more than print the packages being tested - it also prints extra info about the tests being run.

main.go
mvdan: It does look like kingpin enforces a long flag name for all flags. Oh well.

loader/loader.go
mvdan: I think it's going down a slippery slope - shouldl we also print `cat some-file.gunk` any time we read a Gunk file? :)

vishen: Yup, of when we write a file. It looks like Go does some of these, but they likely don't all make sense for Gunk.

main.go
vishen: Ah, gotcha! Done.

main.go
```diff
@@ -30,6 +31,9 @@ func main() {
 }
 
 func main1() int {
+	gen.Flag("print-commands", "print the commands").Short('x').BoolVar(&log.PrintCommands)
+	gen.Flag("verbose", "print the names of packages as they are generated").Short('v').BoolVar(&log.Verbose)
```
vishen: Not sure if these should be here or in an init() function?

mvdan: I don't think there's a big difference for a main package. Here seems fine.

PR #100, opened 2018-11-23
format: add missing proto sequence numbers to a struct

### LINKED ISSUE #74: format: add missing proto field numbers incrementally
Splitting from #37, to keep things tidier.

vishen:
If a struct field is missing the proto sequence number, attempt to add
the sequence number to the tag if it exists, or create a new tag and add
it to that.

However, there is no easy way to update the struct field tag if the 'pb'
tag is already present (whether empty, or containing a non-number). For now,
we return an error to either remove the tag, or add the missing sequence number.

Closes #74


format/format.go
```diff
@@ -123,3 +135,76 @@ func commentFromText(orig ast.Node, text string) *ast.CommentGroup {
...
+		// If there isn't a 'pb' tag present.
+		if !ok {
+			fieldsWithoutSequence = append(fieldsWithoutSequence, f)
+			continue
+		}
+		// If there was a 'pb' tag, but it wasn't empty, return an error.
+		// It is a bit difficult to add in the sequence number if the 'pb'
+		// tag already exists.
+		if ok && val == "" {
+			return fmt.Errorf("struct field tag for pb was empty, please remove or add sequence number")
```
mvdan: these errors should include `filename:line:col` position prefixes

format/format.go
```diff
@@ -123,3 +135,76 @@ func commentFromText(orig ast.Node, text string) *ast.CommentGroup {
...
+		}
+		// If there isn't a number in 'pb' then return an error.
+		i, err := strconv.Atoi(val)
+		if err != nil {
+			return fmt.Errorf("struct field tag for pb contains a non-number %q", val)
```
mvdan: We don't catch this error in `gunk generate`, do we? We probably should too. Perhaps add a TODO.

format/format.go
vishen: How do you get the `filename:line:col`?

vishen: That is very, very, handy! I updated the test to check for this as well.

format/format.go
vishen: Done.

PR #102, opened 2018-11-26
config: remove default config and assume .git is project root

### LINKED ISSUE #101: config: remove default config and assume the project root has either a .git or go.mod
Remove the default config and instead error if there is no `.gunkconfig` found. Currently we assume the project root to be the directory containing a `go.mod` file, we should also check for a `.git/` directory, and if we find one assume that is the project root.

vishen:
Removed the default config. If 'gunk generate' is run and no
'.gunkconfig' is found, an error is now showed to the user. 'gunk
generate' can now only be run if a .gunkconfig file is found.

Previously, we assumed the project root was where a  'go.mod' file was
found. This has been updated to now also look for a '.git' file or
directory.

Closes #101


testdata/scripts/config_noconfig.txt
```diff
@@ -1,28 +1,22 @@
...
-[generate]
-protoc=python
-
 -- gunk/doc.go --
 package util // make this directory a Go package
 
 -- gunk/util.gunk --
 package util
 
+import (
```
vishen: This is showing the wrong diff I believe. I changed `testdata/scripts/config_parent.txt` -> `testdata/scripts/config_projectroot.txt` so I am unsure why it is showing a diff between these two files.

PR #108, opened 2018-11-27
generate: don't include unused imports in the final proto descriptor

### LINKED ISSUE #66: Only include imports that are used when building the proto filedescriptor
We get the following warnings sometimes when using protoc to generate language bindings:

```
echo.proto: warning: Import github.com/gunk/opt/opt.gunk but not used.
echo.proto: warning: Import github.com/gunk/opt/http/http.gunk but not used.
```

This is because we include all imports in the proto filedescriptor, even if they aren't being used. We should remove these imports if they aren't used, although I am unsure how hard this will be.

vishen:
Keep a record of all import paths used for a package. When adding
imports to the proto descriptor check whether that import path has
actually been used by any messages or enums.

Fixes #66


generate/generate.go
```diff
@@ -283,13 +284,21 @@ func (g *Generator) requestForPkg(pkgPath string) *plugin.CodeGeneratorRequest {
 	return req
 }
 
+func (g *Generator) addUsedImport(pkgPath string) {
+	if pkgPath == "" {
```
mvdan: when can this happen and why? please add a comment

generate/generate.go
```diff
@@ -330,7 +339,10 @@ func (g *Generator) translatePkg(pkgPath string) error {
 				// depend on.
 				continue
 			}
-			g.pfile.Dependency = append(g.pfile.Dependency, unifiedProtoFile(opath))
+			// Only include imports that are used.
+			if used, ok := g.usedImports[opath]; ok && used {
```
mvdan: you can shorten this with just `if g.usedImports[opath] {`.

generate/generate.go
vishen: This is just a safe guard for future uses. I will remove it if it is confusing to you though.

mvdan: More than confusing, worrying, Package paths should never be empty, so if we start adding checks for that in the code, we may end up with broken packages that just change gunk's behavior instead of breaking it in obvious ways.

vishen:
Done!

PR #109, opened 2018-11-27
all: allow multiple gunk tags in comments

### LINKED ISSUE #73: Support multiple gunk tags per comment
The code so far only supports zero or one tags per comment, for the sake of simplicity. We can change that quite easily, changing all the code to use a slice instead of a pointer.

vishen:
Changed 'loader.SplitGunkTag' to search through comments for multiple
gunk tags, and will return all found gunk tags for a comment group.

Fixes #73


generate/generate.go
```diff
@@ -344,31 +344,33 @@ func (g *Generator) fileOptions(pkg *loader.GunkPackage) (*desc.FileOptions, err
 		if f.Doc == nil {
 			continue
 		}
-		_, tag, err := loader.SplitGunkTag(g.fset, f.Doc)
-		if err != nil || tag == nil {
+		_, tags, err := loader.SplitGunkTag(g.fset, f.Doc)
+		if err != nil || len(tags) == 0 {
```
mvdan: I imagine this length check is unnecessary, since a range over an empty slice is a noop.

generate/generate.go
```diff
@@ -549,16 +551,18 @@ func (g *Generator) convertService(tspec *ast.TypeSpec) (*desc.ServiceDescriptor
...
-			edesc, val, err := g.interpretTagExpr(tag)
-			if err != nil {
-				return nil, err
-			}
-			if pmethod.Options == nil {
-				pmethod.Options = &desc.MethodOptions{}
-			}
-			if err := proto.SetExtension(pmethod.Options, edesc, val); err != nil {
-				return nil, err
+		if len(tags) > 0 {
```
mvdan: same as before - unnecessary check.

loader/loader.go
```diff
@@ -204,36 +204,53 @@ syntax = "proto3";
...
+func SplitGunkTag(fset *token.FileSet, comment *ast.CommentGroup) (string, []ast.Expr, error) {
+	// Remove the comment leading and / or trailing identifier; // and /* */ and `
 	docLines := strings.Split(comment.Text(), "\n")
-	var tagLines []string
+	var gunkTagLines []string
+	var gunkTagPos []int
+	var commentLines []string
+	// If the +gunk tag is a multiline tag (contains a '{'), then
+	// we need to wait until we find the closing '}'.
+	waitingForMoreLines := false
```
mvdan: I'm not sure I understand why we need this. Why not simply stop each `+gunk` tag at the next `+gunk` line, or at the end of the comments?

Another advantage is that we wouldn't break in cases like:

```
+gunk foo.Bar{
    F: foo.Baz{...}
}
```

loader/loader.go
```diff
@@ -204,36 +204,53 @@ syntax = "proto3";
...
 	for i, line := range docLines {
 		if strings.HasPrefix(line, "+gunk ") {
-			tagLines = docLines[i:]
-			// Replace "+gunk" with spaces, so that we keep the
-			// tag's lines all starting at the same column, for
-			// accurate position information later.
```
mvdan: we don't want to lose this comment.

loader/loader.go
mvdan: A better way to implement your heuristic would be to count opening and closing braces, stopping when we reach 0. Though then we of course have the issue of `+gunk foo.Bar("{")`.

vishen: I went with this approach because I assumed you could have comments in-between gunk tag? If that isn't possible then I agree that your approach is much cleaner.

mvdan: I think not allowing comments in between or after gunk tags is our best short-term strategy. If we want to support those comments in the future, there are ways to do that that would be more reliable than string-based heuristics like the one you implemented. But they'd be tricky to implement and would easily require fifty lines of non-trivial code, so I just don't think it's worth it right now.

loader/loader.go
vishen: Opps, good catch.

generate/generate.go
vishen: Ah, completely missed this! Done.

generate/generate.go

PR #111, opened 2018-11-30
generate: added proto options

vishen:
Added proto options that are defined in 'github.com/gunk/opt'.
Currently, some options aren't defined in the protoc-gen-go
descriptor.pb.go like 'FileOptions.RubyPackage' and
'FileOptions.PhpMetadataNamespace' but they should be there.

Merged 'interpretTagExpr' into 'methodOptions' which will add the http
options straight onto the 'MethodOptions' struct.

Removed leftover cruft 'addUsedImports' from previous PR.


vishen:
I have done this the most trivial way because I am unsure what the changes will look like once the parsing of gunk tags is moved to the loader. @mvdan if you can think of a way to change this to make it easier once the parsing has of gunk tags has moved, let me know I will update the PR :)

generate/proto.go
```diff
@@ -45,6 +46,11 @@ func protoStringOrNil(s string) *string {
 	return proto.String(s)
 }
 
+func protoEnumValue(value constant.Value) int32 {
+	val, _ := constant.Int64Val(value)
```
mvdan: I'd prefer if we do `if !ok { panic(...) }`. Otherwise, we might hide bugs or invalid input since we happily accept those as zero values.

Talking about hypotheticals of course. But I prefer to ensure assumptions via panics, since that means programs will break in less subtle ways if something goes wrong.

testdata/scripts/generate_options.txt
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+    "github.com/gunk/opt/enum"
+    "github.com/gunk/opt/enumvalues"
+    "github.com/gunk/opt/file"
+    "github.com/gunk/opt/file/java"
+    "github.com/gunk/opt/field"
+    "github.com/gunk/opt/message"
+    "github.com/gunk/opt/service"
+    "github.com/gunk/opt/method"
```
mvdan: please use tabs in all these

generate/proto.go
vishen: My understanding was that this would get picked up in the `constant.Int64Val` function, it panics if the value is not an int, and since the gunk tags are type checked there couldn't be a situation where the type is anything but an int? I have added the panic for safe keeping regardless.

testdata/scripts/generate_options.txt
vishen: Done!

generate/proto.go
mvdan: Oh, the `!ok` happens if the value isn't exact, not if it's not an integer. A value can be non-exact if it's unknown because the program had a type error somewhere. So I do think we should check, just in case.

generate/proto.go
```diff
@@ -45,6 +46,14 @@ func protoStringOrNil(s string) *string {
 	return proto.String(s)
 }
 
+func protoEnumValue(value constant.Value) int32 {
+	val, ok := constant.Int64Val(value)
+	if !ok {
+		panic(fmt.Sprintf("value %#v was not an int", value))
```
mvdan: this should now be amended - say "was not exact".

> If x is Unknown, the result is (0, false).

mvdan: Err, I guess a better way would be "value ... is unknown"

generate/proto.go
vishen: Ah, gotcha! Updated wording.

generate/proto.go
vishen: Done!

PR #118, opened 2018-12-11
all: bundle some proto libraries with gunk

### LINKED ISSUE #116: Java fails when using google api annotations.
When using google api annotations (`http.Match`), we are not adding the `google/api/annotations.proto` as a dependency to the `FileDescriptorProto`. This is fine for all proto generators, except for Java which errors because the dependency is missing (even though it doesn't use the custom extensions). 

This is the error:
```
protoc --java_out=/home/jonathan/src/gunk/examples/util/v1 --descriptor_set_in=/dev/stdin all.proto
[libprotobuf FATAL ../../../../../src/google/protobuf/compiler/java/java_file.cc:129] CHECK failed: file_proto_desc: Find unknown fields in FileDescriptorProto when building all.proto. It's likely that those fields are custom options, however, descriptor.proto is not in the transitive dependencies. This normally should not happen. Please report a bug.
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: file_proto_desc: Find unknown fields in FileDescriptorProto when building all.proto. It's likely that those fields are custom options, however, descriptor.proto is not in the transitive dependencies. This normally should not happen. Please report a bug.
error: error executing "protoc": "", signal: aborted (core dumped)
```

I can fix the error by loading and adding `google/api/annotations.proto` to the `FileDescriptorProto`, but I am unsure if we need to do this, or this is a bug in protoc.

I have raised an issue with protoc https://github.com/protocolbuffers/protobuf/issues/5415 regarding this inconsistency. Depending on whether the issue is with the java generator, or with all the other generators, we may have to add the dependency for google api annotations (or any other custom extension) when creating the `FileDescriptorProto`.

vishen:
Bundle the following proto libraries with Gunk:

- google/api/annotations.proto
- google/protobuf/empty.proto
- google/protobuf/timestamp.proto
- google/protobuf/duration.proto

these are the libraries that are currently used by Gunk, and require
calling out to protoc to load from disk. These will now be included in
Gunk.

The included libraries are being stored in their FileDescriptorProto
on-the-wire format. Go generate is used to create these on-the-wire
formats of the libraries, which runs 'protoc' to load the libraries from
disk and write the on-the-wire formats to the 'generated/' directory.

Currently using 'shurcooL/vfsgen' to create gzipped contents of the
files and store them in memory.

Fixes #116


vishen:
For me, this approach was actually much easier than the one you described. I initially attempted the approach you described but ran into hurdles (although I may have been approaching it incorrectly), where-as this approach was surprisingly quite straightforward. Also, it seemed logical to me to try and eliminate an extra call out to `protoc` if it isn't required. Happy to try and make the other way work if you think that way is better for whatever reason? :)

mvdan:
Ah, if this was simpler, then sure. I'm only talking in terms of what I'd try, I haven't actually tried :)

I presume we'll refactor this again at some point soon, so I don't want to worry too much about it now.

vishen:
Yup, I figured we will likely throw this out when the issue is fixed with protoc or when we start downloading `protoc` :)

PR #127, opened 2019-01-02
generate: add missing type mappings

### LINKED ISSUE #115: generate: add missing Go type to proto type mappings.
Currently, we only have Go basic types `int`, `int32`, `bool` and `string` mapping to proto types (https://github.com/gunk/gunk/blob/master/generate/generate.go#L1001). We should add the rest of the mappings:

Go types: https://godoc.org/go/types#BasicKind
Proto types: https://godoc.org/github.com/gogo/protobuf/protoc-gen-gogo/descriptor#FieldDescriptorProto_Type

- uint32 -> TYPE_UINT32
- int64 -> TYPE_INT64
- uint64 -> TYPE_UINT64
- float32 -> TYPE_FLOAT
- float64 -> TYPE_DOUBLE

I have likely missed some, but these are the main ones I can see.

vishen:
Added the following missing type conversions from gunk (go) -> proto:

- uint -> FieldDescriptorProto_TYPE_UINT32
- uint32 -> FieldDescriptorProto_TYPE_UINT32
- int64 -> FieldDescriptorProto_TYPE_INT64
- uint64 -> FieldDescriptorProto_TYPE_UINT64
- float32 -> FieldDescriptorProto_TYPE_FLOAT
- float64 -> FieldDescriptorProto_TYPE_DOUBLE

Fixes #115

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -895,13 +895,23 @@ func (g *Generator) qualifiedTypeName(typeName string, pkg *types.Package) strin
 func (g *Generator) convertType(typ types.Type) (desc.FieldDescriptorProto_Type, desc.FieldDescriptorProto_Label, string) {
 	switch typ := typ.(type) {
 	case *types.Basic:
-		// Map Go types to proto types via
+		// Map Go types to proto types:
 		// https://developers.google.com/protocol-buffers/docs/proto3#scalar
 		switch typ.Kind() {
 		case types.String:
 			return desc.FieldDescriptorProto_TYPE_STRING, desc.FieldDescriptorProto_LABEL_OPTIONAL, ""
 		case types.Int, types.Int32:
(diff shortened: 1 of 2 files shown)


generate/generate.go
```diff
@@ -895,13 +895,23 @@ func (g *Generator) qualifiedTypeName(typeName string, pkg *types.Package) strin
 func (g *Generator) convertType(typ types.Type) (desc.FieldDescriptorProto_Type, desc.FieldDescriptorProto_Label, string) {
 	switch typ := typ.(type) {
 	case *types.Basic:
-		// Map Go types to proto types via
+		// Map Go types to proto types
```
mvdan: If you want to split these two lines into two sentences, add some punctuation, like `:`.

PR #128, opened 2019-01-02
config: allow shortened generate sections for protoc-gen-*

### LINKED ISSUE #96: config: shortened generate stanza is only for protoc
Our shortened generate stanza in a gunkconfg file is currently always a short-hard for the `protoc` command:

```
[generate js]

[generate go]
plugins=grpc
```

is short-hand for

```
[generate]
protoc=js

[generate]
protoc=go
plugins=grpc
```

which will invoke protoc like the following; `protoc --js_out` and `protoc --go_out=pluings=grpc`. We will likely want the shortened generate to work for both the protoc command and protoc-gen-* commands.

One way would be to check what `x` is in `[generate x]` and if it is a language that protoc can generate code for (https://github.com/protocolbuffers/protobuf#protobuf-runtime-installation), we expand that to be:

```
[generate]
protoc=<x>
```

but if it isn't a language that protoc can generate code for, we assume that is it a protoc-gen-* plugin, and expand it to

```
[generate]
command=protoc-gen-<x>
```

vishen:
'protoc-gen-*' generators can now be used as shortened generate sections
in the gunkconfig. Previously, we just assumed that any
'[generate <x>]' in the gunkconfig should be passed through as 'proto --<x>_out'.
Now, we can have '[generate go]' and this will check for a 'protoc-gen-go' binary,
if it exsists that is used, otherwise it falls back to being passed
through to protoc.

Fixes #96

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -180,14 +181,22 @@ func load(reader io.Reader) (*Config, error) {
...
-			// Check to see if we have the shorten version of a generate config:
-			// [generate js].
+
 			gen, err = handleGenerate(s)
-			gen.ProtocGen = strings.Trim(sParts[1], "\"")
+			generator := strings.Trim(sParts[1], "\"")
+			// Is this shortened generator a protoc-gen-* binary, or
+			// should it be passed to protoc.
+			if _, err := exec.LookPath("protoc-gen-" + generator); err == nil {
+				gen.Command = "protoc-gen-" + generator
```
mvdan: why not use the found path directly, since we do the work for that here already?

the only disadvantage I can see is that if we support `-x` for generators, then we'd show `/bin/some-generator flags...` instead of `some-generator flags...`. If that's the reason, seems fine to me to keep it as it is, but I'd add a comment to prevent others from slightly optimising the code like I suggested.

vishen: We can use the path directly, but I wanted to keep it the same as the normal generate section that doesn't lookup the path. If we wanted to use the path we should also do the same for the normal generate section.

mvdan: True. Please add a short comment :)

vishen: Done.

PR #135, opened 2019-01-04
docs: update readme

### LINKED ISSUE #125: Running Gunk without Go installed
I was trying to use Gunk in a Docker container without Go installed and ran into the following error:

```
/examples/util # gunk generate -v -x .
error: couldn't exec 'go [list -e -json -compiled -test=false -export=false -deps=false -find=true -- .]': exec: "go": executable file not found in $PATH *exec.Error
```

If we need to have a hard dependency on Go, I think we may need to add something to the documentation about requiring Go installed?

jwangsadinata:
Closes #125.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![GoDoc][card-godoc]][godoc]
+[![Build Status][card-travis]][travis]
 
 # Gunk
 
@@ -107,7 +108,9 @@ comes with Go 1.11 and above. After adding a new dependency, please run the foll
     $ GO111MODULE=on go mod tidy
 
 [card-godoc]: https://godoc.org/github.com/gunk/gunk?status.svg
+[card-travis]: https://travis-ci.org/gunk/gunk.svg?branch=master
 [godoc]: https://godoc.org/github.com/gunk/gunk
 [go-modules]: https://github.com/golang/go/wiki/Modules
 [go-project]: https://golang.org/project
 [issue-tracker]: https://github.com/gunk/gunk/issues
+[travis]: https://travis-ci.org/gunk/gunk
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-[![GoDoc][card-godoc]][godoc]
-[![Build Status][card-travis]][travis]
+[![GoDoc](https://godoc.org/github.com/gunk/gunk?status.svg)](https://godoc.org/github.com/gunk/gunk)
+[![Build Status](https://travis-ci.org/gunk/gunk.svg?branch=master)](https://travis-ci.org/gunk/gunk)
 
 # Gunk
 
-Gunk is an acronym for "Gunk Unified N-terface Kompiler."
+Gunk is an acronym for "Gunk Unified N-terface Kompiler".
 
 Gunk primarily works as a frontend for `protobuf`'s, `protoc` compiler. `gunk`
 aims is to provide a way to work with `protobuf` files / projects in the same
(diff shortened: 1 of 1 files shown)


jwangsadinata:
@mvdan, @vishen, let me know what should be put on usage or examples, that might be beneficial to whoever is reading this. I worked on adding some structure/cosmetics to the readme, but not so much yet on the actual usage. I can work on the basic ones, but you guys might have better ideas in mind.

README.md
```diff
@@ -1,4 +1,7 @@
-# About gunk
+[![GoDoc][card-godoc]][godoc]
+[![Build Status][card-travis]][travis]
```
mvdan: Wow, what kind of black magic is this? I always thought one needed to include the full link for these to work.

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+* [Installing](#installing)
+* [Usage](#usage)
+* [Contributing](#contributing)
+
+## How it works
+
+Gunk works by using Go's standard library `go/ast` package (and related
+packages) to parse all `*.gunk` files in a "Gunk Package" (and applying the
+same kinds of rules as Go package), building the appropos protobuf messages. In
+turn, those are passed to `protoc-gen-*` tools, along with any passed options.
```
mvdan: I think this shows a bit too much internal detail to be this far up in the README. Users will spend 30s skimming the readme to read what this is, why they should use it, and how to install it - the implementation details are less important.

So I'd just move it at least below "usage".

README.md
mvdan: Reading the end of the markdown, it now dawns on me. I don't think it makes sense to separate these links. These are single-line badges - I personally think that `[links]` only make sense within paragraphs, to keep them more readable.

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+
+## Features
+
+- Written in simple and idiomatic [Go][go-project]
+- Intuitive and [easy to use](#usage)
+- Works on Linux, Mac and Windows
```
mvdan: Seems to me like this section doesn't add a lot of meat to the README. The first point repeats what the first few paragraphs say. The second just links to "usage". The third does add extra info, but we could just add system requirements to the installation section.

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
 ## Installing
 
-Gunk can be installed in the usual Go fashion:
+Gunk requires [Go][go-project] to be installed on your machine. To install Go,
+follow the [installation instructions][go-install] for your specific operating
+systems.
+
+Gunk can then be installed in the usual Go fashion:
+
+	$ go get -u github.com/gunk/gunk
```
mvdan: I personally slightly prefer leaving out the `$` when we're just giving commands for people to run. Otherwise copy-pasting is a bit more difficult, as one has to be careful not to copy the dollar.

I do use dollars when showcasing how something works though, to differentiate the prompt from the output:

```
$ gunk awesome-stuff
Wow such awesome stdout!
```

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+
+  generate [<flags>] [<patterns>...]
+    Generate code from Gunk packages.
+
+  convert [<flags>] [<file>]
+    Convert Proto file to Gunk file.
+
+  format [<patterns>...]
+    Format Gunk code.
+
```
mvdan: remove the empty line please

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+```
+
+This project uses [go modules][go-modules] for managing dependencies, which
+comes with Go 1.11 and above. After adding a new dependency, please run the following:
```
mvdan: `tidy` should be run after removing or changing deps too, in case anything needs pruning. I'd just say "after making any changes to the imported packages" or "before sending a PR".

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+### Bug Reports & Feature Requests
+
+Please use the [issue tracker][issue-tracker] to report any bugs or file feature requests.
+
+### Developing
+
+PRs are welcome. To begin developing, do this:
+
+```bash
+$ git clone git@github.com:gunk/gunk.git
```
mvdan: https please - otherwise you force the user to have an SSH key set up with a github account.

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
+$ export GO111MODULE=on
+$ cd gunk/
+$ go build -o gunk
```
mvdan: 99% sure you can omit `-o gunk` here

README.md
```diff
@@ -12,30 +15,107 @@ is simpler and easier to work with. Additionally, for developers familar with
...
 
-## Overview
+### More information
 
-Gunk works by using Go's standard library `go/ast` package (and related
-packages) to parse all `*.gunk` files in a "Gunk Package" (and applying the
-same kinds of rules as Go package), building the appropos protobuf messages. In
-turn, those are passed to `protoc-gen-*` tools, along with any passed options.
+Please see [the GoDoc API page][godoc] for a
+full API listing.
```
mvdan: Please keep a somewhat consistent column limit :)

mvdan:
LGTM, but please remove the WIP part so we can merge. Unless you mean to leave this open?

jwangsadinata:
right, sorry about that. it's done now @mvdan.

PR #139, opened 2019-01-08
generate: sort protofile dependencies

### LINKED ISSUE #137: generate: proto file order needs to have dependencies listed first for grpc_tools_node_protoc_plugin
When trying to generate grpc code with `grpc_tools_node_protoc_plugin` for nodejs, it currently fails:

```
# simple.gunk
package simple

type Message struct {
    Msg string `pb:"1"`
}

type SimpleService interface {
    Handle() Message
}
```
```
# .gunkconfig
[generate]
command=/usr/bin/grpc_tools_node_protoc_plugin
```
```
$ gunk generate ./... -v -x                                                                                                                                                                  
/usr/bin/grpc_tools_node_protoc_plugin 
[libprotobuf ERROR google/protobuf/descriptor.cc:3457] Invalid proto descriptor for file "_/home/jonathan/src/gunk-tools/loading/gunk/all.proto":
[libprotobuf ERROR google/protobuf/descriptor.cc:3460]   _/home/jonathan/src/gunk-tools/loading/gunk/all.proto: Import "google/protobuf/empty.proto" has not been loaded.
[libprotobuf ERROR google/protobuf/descriptor.cc:3460]   simple_gunk.SimpleService.Handle: ".google.protobuf.Empty" is not defined.
/usr/lib/node_modules/grpc-tools/bin/protoc_plugin.js:37
    throw error;
```

Gunk's ordering of the `ProtoFile` list:

```
- package_name=all.proto -> dependencies=[google/protobuf/empty.proto]
- package_name=google/protobuf/empty.proto -> dependencies=[]
```

Protoc's ordering of the `ProtoFile` list:

```
- package_name=google/protobuf/empty.proto -> dependencies=[]
- package_name=all.proto -> dependencies=[google/protobuf/empty.proto]
```

Changing the order in Gunk to be the same as Protoc seems to solve the issue.

My best guess is  because `grpc_tools_node_protoc_plugin` expects that any `ProtoFile` with dependencies has those dependencies declared first in the list. Currently we just order based on the `ProtoFile.Name`, but we need to alter this to have a ProtoFile's dependencies declared earlier in the `ProtoFile` list.

`grpc-gateway` also requires the `ProtoFile` list to be in some order, so we need to make sure that still works with any changes we make to solve this issue.

vishen:
Attempt to sort the FileDescriptorSet in an order such that any proto
files without dependencies are first in the list, and the rest are at
the end.

Use a fairly dumb, but simple approach to determine which protofile's
depend on others. We can use this approach because we don't load that
many protofile's, and the ones we do load don't depend on each other.

This is required to fix an issue with the protoc nodejs grpc generator,
which seems to expect the protofiles to be in an order where any
depdendent protofile's are earlier on in the list.

Fixes #137

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -279,6 +279,20 @@ func (g *Generator) requestForPkg(pkgPath string) *plugin.CodeGeneratorRequest {
 		return *req.ProtoFile[i].Name < *req.ProtoFile[j].Name
 	})
 
+	// Super dumb approach to get which ProtoFile's depend on
+	// other ProtoFiles. This works because we don't load
+	// many protofiles at the moment and the ones we do load
+	// don't depend on each other.
+	independentProtoFiles := make([]*desc.FileDescriptorProto, 0, len(req.GetProtoFile()))
+	dependentProtoFiles := make([]*desc.FileDescriptorProto, 0, len(req.GetProtoFile()))
+	for _, f := range req.GetProtoFile() {
+		if len(f.GetDependency()) == 0 {
+			independentProtoFiles = append(independentProtoFiles, f)
+		} else {
+			dependentProtoFiles = append(dependentProtoFiles, f)
+		}
+	}
+	req.ProtoFile = append(independentProtoFiles, dependentProtoFiles...)
 	return req
 }
 


vishen:
@mvdan I am not sure if you have a better approach in mind? I went with the simplest for now, but happy to change if you have other ideas.

vishen:
I am unsure how to add a test for this, the plugin was a node-based protoc plugin and short of introducing that into the test system I am not sure how else to test this. We could start introducing unit-tests?

mvdan:
Unit test sounds good to me then. You could have the test iterate over the file descriptors, and check that all dependencies are satisfied by earlier file descriptors.

While at it, we could make the same unit test check that there are no cyclic dependencies. We could call it something like `TestStrictFileDescriptorSet` or something, and put in there all the weird constraints and assumptions we've seen generators enforce on the file set.

mvdan:
Also, I know how to write the above, so I can do that tomorrow along with my refactor of your code. I'll just ask you to run the node generator again, just in case I got the test wrong :)

PR #147, opened 2019-01-10
convert: allow proto files and folders

### LINKED ISSUE #134: Gunk convert should support directories
Current `gunk convert` only accepts a single proto file to convert, this should be changed to accept a directory as well and convert any proto files found in there to Gunk files.

We may also want to add an output path configuration.

vishen:
Accept a list of files and folders for the convert command and run
these through the converter. Only '.proto' files are converted to Gunk
files.

Updated readme.

Update github.com/emicklei/proto to 1.6.8.

Fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,12 +185,12 @@ binary
 
 ## Convert
 
-Gunk provides functionality to convert a proto file to a Gunk file.
+Gunk provides functionality to convert proto files or folders to a Gunk file.
 
-    gunk convert path/to/file.proto
+    gunk convert path/to/file.proto other/path/to/file.proto
+    gunk convert ./path/to/proto/dir
 
-This will convert your proto file to the equivalent Gunk file. Currently
-this only works for single proto files.
+This will convert your proto file to the equivalent Gunk file.
 
 ## Gunk Annotations
 
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -35,9 +35,49 @@ type builder struct {
 	imports []*proto.Import
 }
 
-// Run converts a proto file to a gunk file, saving the file in the same folder
-// as the proto file.
-func Run(path string, overwrite bool) error {
+// Run converts proto files or folders to gunk files, saving the files in
+// the same folder as the proto file.
+func Run(paths []string, overwrite bool) error {
+	for _, path := range paths {
+		if err := run(path, overwrite); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func run(path string, overwrite bool) error {
+
+	fi, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	// Determine whether the path is a file or a directory.
+	// If it is a file convert the file.
+	if !fi.IsDir() {
+		return convertFile(path, overwrite)
+	} else if filepath.Ext(path) == ".proto" {
+		// If the path is a directory and has a .proto extension then error.
+		return fmt.Errorf("%s is a directory, should be a proto file.", path)
+	}
+
+	// Handle the case where it is a directory. Loop through
+	// the files and if we have a .proto file attempt to
+	// convert it.
+	files, err := ioutil.ReadDir(path)
+	for _, f := range files {
+		// If the file is not a .proto file
+		if f.IsDir() || filepath.Ext(f.Name()) != ".proto" {
+			continue
+		}
+		if err := convertFile(filepath.Join(path, f.Name()), overwrite); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func convertFile(path string, overwrite bool) error {
 	if filepath.Ext(path) != ".proto" {
 		return fmt.Errorf("convert requires a .proto file")
 	}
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -185,9 +185,10 @@ binary
...
 ## Convert
 
-Gunk provides functionality to convert a proto file to a Gunk file.
+Gunk provides functionality to convert proto files or folders to a Gunk file.
 
-    gunk convert path/to/file.proto
+    gunk convert path/to/file.proto other/path/to/file.proto
+    gunk convert ./path/to/proto/files
 
 This will convert your proto file to the equivalent Gunk file. Currently
```
jwangsadinata: delete the `Currently this only works for single proto files` sentence?

convert/convert.go
```diff
@@ -35,9 +35,45 @@ type builder struct {
 	imports []*proto.Import
 }
 
-// Run converts a proto file to a gunk file, saving the file in the same folder
-// as the proto file.
-func Run(path string, overwrite bool) error {
+// Run converts proto files or folders to a gunk files, saving the file in
```
jwangsadinata: nit: `s/to a gunk files, saving the file/to gunk files, saving the files/`

README.md
vishen: Ah oops, good catch!

convert/convert.go
vishen: Done.

README.md
```diff
@@ -185,12 +185,12 @@ binary
 
 ## Convert
 
-Gunk provides functionality to convert a proto file to a Gunk file.
+Gunk provides functionality to convert proto files or folders to a Gunk file.
 
-    gunk convert path/to/file.proto
+    gunk convert path/to/file.proto other/path/to/file.proto
+    gunk convert ./path/to/proto/files
```
mvdan: I'd be more explicit, e.g. `./path/to/proto/dir`

convert/convert.go
```diff
@@ -35,9 +35,45 @@ type builder struct {
...
+	for _, path := range paths {
+		if err := run(path, overwrite); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func run(path string, overwrite bool) error {
+	fi, err := os.Stat(path)
```
mvdan: you might want to make this logic simpler. for example, the way `go build` does it, it builds files if the first one has a `.go` extension. that way, `go build ./foo.go` errors if `foo.go` is something very weird like a directory, and they save the extra stat work. you likely also want to error on cases like `gunk convert ./some-folder` where `some-folder` is actually a proto file.

convert/convert.go
```diff
@@ -90,6 +126,8 @@ func Run(path string, overwrite bool) error {
 	// TODO: We should run this through the Gunk generator to
 	// make sure that it compiles?
 
+	fmt.Printf("%s", w.String())
```
mvdan: delete?

convert/convert.go
vishen: I don't quite follow sorry? I also don't see how this logic could be simpler, it is finding all the .proto files in a directory or using the path if it is a .proto file?

mvdan: Right now you do:

```
if !IsDir(path) {
    return convertFile(...)
}
// convert all files within dir that end in ".proto"
```

What I was suggesting was to instead do:

```
if HasSuffix(path, ".proto") {
    return convertFile(...)
}
// convert all files within dir that end in ".proto"
```

This is the same thing that `go build` does. For example, try doing `go build foo.txt` if the file exists. It won't error saying that the file has bad syntax - it will say that it can't find the package, because the argument isn't treated as a Go file.

vishen: What if `path` is a directory in this case? The error in the case where we take your approach is:

```
error: unable to parse proto file "some.proto": go scanner error at <input> = read some.proto: is a directory
```

Where if we check that the `path` is a directory then we can correctly handle the case where someone is using a `example.proto` as a directory name.

mvdan: Yeah, Go doesn't support directories with `.go` extensions, and I don't think we should support that for gunk or proto directories either. In their case, they do catch that error to give a slightly better one:

```
$ mkdir foo.go
$ go build foo.go
foo.go is a directory, should be a Go file
```

So I don't think this would save you the stat call, then :)

vishen: Is there a reason not to support that though? I can't think of a good reason not to? It seems like if someone wants to name their directories like that, and it is trivial for us to handle, then should we handle this case?

Happy to remove it if there is a reason that this is a bad idea though :)

mvdan: I've always assumed that the reason why they keep the logic string-based is because when one sees `go build foo.go` or `go build ./bar`, it's always clear whether we're building a single file or a directory/package. It's not clear if it depends on what's on disk. This is why I was initially saying "simpler logic".

Ultimately we've got bigger fish to fry, so feel free to go either way :)

vishen: Ah, seems reasonable to me actually. If someone actually hits this issue we can always revisit the way we do this. I have added a check for `IsDir(path) && HasSuffix(path, ".proto")` to error on this case.

PR #151, opened 2019-01-15
config: global 'out' parameter

vishen:
Allow a global 'out' parameter to be set. Each '[generate]' section will
inherit this global 'out' parameter unless the '[generate]' section has
the 'out' parameter set.

### DIFF
--- a/testdata/scripts/config_global.txt
+++ b/testdata/scripts/config_global.txt
@@ -1,0 +1,38 @@
+env HOME=$WORK/home
+
+gunk generate ./gunk
+
+# Check that that files were generated in the correct folder
+exists gunk/v1/all.pb.go gunk/v2/all_pb2.py
+
+-- go.mod --
+module testdata.tld/util
+
+-- gunk/v1/.empty --
+-- gunk/v2/.empty --
+
+-- gunk/.gunkconfig --
+# Set global out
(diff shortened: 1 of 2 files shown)


testdata/scripts/config_global.txt
```diff
@@ -0,0 +1,38 @@
...
+gunk generate ./gunk
+
+# Check that that files were generated in the correct folder
+exists gunk/v1/all.pb.go gunk/v2/all_pb2.py
+
+-- go.mod --
+module testdata.tld/util
+
+-- gunk/v1/.empty --
+-- gunk/v2/.empty --
```
mvdan: why do we need these? if gunk is not creating the directories, I think it should be.

vishen: Yup, `gunk generate` currently behaves similar to `protoc` which doesn't create the directories for you (which is why I think we went with this). If you think Gunk should do this, we should raise an issue.

mvdan: Ah, being consistent with protoc makes sense I guess. I find that weird of protoc, but at least we're being consistent.

Let's see if users complain about this, or if it's a problem in practice. I personally find it slightly annoying, but it's not a big deal.

Issue #143, opened 2019-01-08
convert: convert proto options to gunk options


vishen:
At the moment we either ignore or error on all proto options when converting proto to gunk. We should attempt to convert the proto options to gunk options if possible.

PR #155, opened 2019-01-16
convert: proto fileoptions

vishen:
Updated convert to convert proto 'FileOptions' to the corresponding Gunk
annotations.

While doing this, it was required to keep a record of which
imports of Gunk annotations are being used. Now we only output imports
that are actually in use, or no imports at all if there are no imports.

Updates #143

### DIFF
--- a/testdata/scripts/convert_fileoptions.txt
+++ b/testdata/scripts/convert_fileoptions.txt
@@ -1,0 +1,25 @@
+env HOME=$WORK/home
+
+gunk convert util.proto
+cmp util.gunk util.gunk.golden
+
+-- util.proto --
+syntax = "proto3";
+
+package util;
+
+option optimize_for = 2;
+
+option java_package = "com.example.util"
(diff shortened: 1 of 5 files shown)


testdata/scripts/convert_fileoptions.txt
```diff
@@ -0,0 +1,25 @@
...
+
+package util;
+
+option optimize_for = 2;
+
+option java_package = "com.example.util"
+option java_outer_classname = "JavaOuterClassname"
+
+-- util.gunk.golden --
+// +gunk file.OptimizeFor(2)
```
vishen: @mvdan If we wanted to output `file.OptimizeFor(file.CodeSize)` (https://github.com/gunk/opt/blob/master/file/file.gunk#L12), do you whether that is easy or not?

mvdan: Should be reasonable to do. I'll have a look.

PR #164, opened 2019-01-18
Fix godoc/build status links in README.md

Gnouc:


kenshaw:
Fixed via a different way.

Issue #182, opened 2019-01-21
gunk format should fix the incorrect pb numbering


jwangsadinata:
I recently used `gunk` to generate the artifacts for my gunk file, and it did not return any error of my malformed gunk file. The pb-numbering was incorrectly defined like the following:

```go
type A struct {
    ID     string `pb:"1" json:"id"`
    Name   string `pb:"1" json:"name"`
    B      string `pb:"2" json:"b"`
    C      uint   `pb:"3" json:"c"`
    D      uint   `pb:"4" json:"d"`
    E      uint   `pb:"5" json:"e"`
}
```

and the generated code looks like this, which may result in the wrong formatting of data:
```go
type A struct {
    ID                   string   `protobuf:"bytes,1,opt,name=ID,json=id,proto3" json:"ID,omitempty"`
    Name                 string   `protobuf:"bytes,1,opt,name=Name,json=name,proto3" json:"Name,omitempty"`
    B                    string   `protobuf:"bytes,2,opt,name=B,json=b,proto3" json:"B,omitempty"`
    C                    uint32   `protobuf:"varint,3,opt,name=C,json=c,proto3" json:"C,omitempty"`
    D                    uint32   `protobuf:"varint,4,opt,name=D,json=d,proto3" json:"D,omitempty"`
    E                    uint32   `protobuf:"varint,5,opt,name=E,json=e,proto3" json:"E,omitempty"`
    XXX_NoUnkeyedLiteral struct{} `json:"-"`
    XXX_unrecognized     []byte   `json:"-"`
    XXX_sizecache        int32    `json:"-"`
}
```

Re-running `gunk generate` on a slightly different proto and directory structure, I found this:
```bash
$ gunk generate .
error: error executing "protoc": exit status 1: all.proto: Field number 1 has already been used in "pokemon.Pokemon" by field "Name".
all.proto: No such file or directory
```

but running `gunk format` does not fix the issues at all.

The suggestion is to implement pb-numbering checks on `gunk format` so that issues like this can be trivially solved by using it.

PR #184, opened 2019-01-21
loader: error on repeated sequence numbers

vishen:
If we detect the same sequence number used in a struct, return an error
pointing to the duplicated sequence number.

Fixes: #182

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -1,6 +1,7 @@
 package loader
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"go/ast"
@@ -16,6 +17,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
@@ -27,6 +29,23 @@ import (
 	"github.com/gunk/gunk/assets"
 	"github.com/gunk/gunk/log"
 )
+
+type ValidateError struct {
+	errors []error
+}
+
+func (ve ValidateError) Error() string {
+	var buf bytes.Buffer
+	fmt.Fprintf(&buf, "got %d errors:\n", len(ve.errors))
+	for i, err := range ve.errors {
+		fmt.Fprintf(&buf, "%v", err)
+		if i < len(ve.errors)-1 {
+			fmt.Fprintf(&buf, "\n")
+		}
+	}
+	return buf.String()
+
+}
 
 type Loader struct {
 	Dir  string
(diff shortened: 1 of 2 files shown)


vishen:
@mvdan I have mode that check for duplicate sequence numbers to the loader. Since `generate` and  `format` treat some error cases differently we can't have all error / sanity checking in the loader.

loader/loader.go
```diff
@@ -170,16 +171,80 @@ func (l *Loader) Load(patterns ...string) ([]*GunkPackage, error) {
...
+				formatErr = ierr.err
+			} else {
+				panic(r)
+			}
+		}
+	}()
+	for _, file := range gunkPackage.GunkSyntax {
+		ast.Inspect(file, func(node ast.Node) bool {
+			switch node.(type) {
+			case *ast.StructType:
```
mvdan: use an if-not-ok-early-return and unindent  the rest of the code

loader/loader.go
```diff
@@ -170,16 +171,80 @@ func (l *Loader) Load(patterns ...string) ([]*GunkPackage, error) {
...
+				st := node.(*ast.StructType)
+				if st.Fields == nil {
+					return true
+				}
+
+				// Check for struct tag 'pb' and ensure that if it does exist
+				// it is a valid integer, and it is unique in that struct.
+				// The other validation should happen in format and generate
+				// as they both treat the same error cases differently.
+				usedSequences := []int{}
```
mvdan: perhaps I'm missing something obvious, but a `map[int]bool` would be way simpler here. you can argue that it's overkill since the lengths will always be tiny, but the code would be way simpler and we're not even benchmarking/optimising.

loader/loader.go
```diff
@@ -170,16 +171,80 @@ func (l *Loader) Load(patterns ...string) ([]*GunkPackage, error) {
...
 		pkgs = append(pkgs, pkg)
 	}
-
 	return pkgs, nil
 }
 
+// validatePackage will validate that a given GunkPackage. It should
+// check sanity checks and common error cases that are shared with
+// the format and generate packages.
+func validatePackage(gunkPackage *GunkPackage) (formatErr error) {
```
mvdan: just a thought - perhaps this could return a `[]error`. it depends on how easy it is for these validity checks to continue after finding one error. not a problem either way, but it would be easier for the human if the tool gave many errors at once, just like most compilers do.

loader/loader.go
vishen: Done.

loader/loader.go
vishen: Done.

loader/loader.go
vishen: Done.

PR #188, opened 2019-01-21
generate: greatly simplify file descriptor sets

mvdan:
Generators directly within protoc insist on writing the files
themselves, so we must give it a CodeGeneratorRequest that has been
modified to contain the right paths.

However, we were doing this in a very roundabout way. We did it for all
generators, and we used a map, which is unnecessary.

The actual logic only requires about ten lines of code in
generateProtoc. We can remove over fifty lines, thanks to that.

The other piece that the old code did was replace ".gunk" with ".proto".
However, all packages are generated as a single "all.proto" file now, so
that line had been a no-op for a while.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -128,7 +128,8 @@ func FileDescriptorSet(dir string, args ...string) (*desc.FileDescriptorSet, err
 
 	// Generate the filedescriptorset for the Gunk package.
 	req := g.requestForPkg(pkgs[0].PkgPath)
-	return g.generateFileDescriptorSet(*req), nil
+	fds := &desc.FileDescriptorSet{File: req.ProtoFile}
+	return fds, nil
 }
 
 type Generator struct {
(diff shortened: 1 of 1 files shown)


generate/generate.go
```diff
@@ -128,7 +128,8 @@ func FileDescriptorSet(dir string, args ...string) (*desc.FileDescriptorSet, err
 
 	// Generate the filedescriptorset for the Gunk package.
 	req := g.requestForPkg(pkgs[0].PkgPath)
-	return g.generateFileDescriptorSet(*req), nil
+	fds := &desc.FileDescriptorSet{File: req.ProtoFile}
```
vishen: We should do the same post-processing here that we do in `generateProtoc()`. Ideally when we `dump` the `FileDescriptorSet`, it should be the exact same as what we are sending to `protoc`. If we were to pass in this `FileDescriptorSet` to `protoc` it wouldn't work the same as the one in `generateProtoc()`

mvdan: Is that how `gunk dump` is supposed to work, though? I always assumed it was for debugging purposes, or to reuse it for any given purpose - not exclusively to feed it to `protoc`.

For example, for the purpose of debugging, it's best if we leave the filenames untouched. `some/package/all.proto` is more descriptive than just `all.proto`.

Unless you do use `gunk dump` with protoc often? Seems like protoc requires other flags and parameters to do its thing, so I presume `dump`'s output can rarely be enough on its own.

vishen: That's how I have used it in the past, feed it into `protoc` to make sure it works as expected.

I do agree that it is more readable this way, but I would be a bit confused that I couldn't feed it into `protoc` to see what the actual output of `protoc` is.

Maybe we can have another flag or format to dump that will produce a `FileDescriptorSet` that is the same as what we send to `protoc`? Although we can raise this as an issue and handle it in another PR.

mvdan: Well, it should work either way. It's just that with this PR as-is, `protoc` would write to the path `some/package/all.proto` instead of `all.proto`. Which I'd even say is better, since otherwise you can't generate multiple packages at once easily - they'd all try to write to the same directory :)

If you find ways to make `dump` easier to use for your needs, we can discuss that in future issues and PRs. But I don't think it's necessarily a good idea to force it to do what is convenient for protoc, at least by default.

PR #189, opened 2019-01-23
loader: rethink package loading errors

mvdan:
go/packages only uses its error return for fatal errors, such as when
the query is malformed or when it cannot call 'go list'.

All other errors, such as typecheck errors, are attached to packages.
This is because loading can often continue in the face of these
non-fatal errors, and it's useful for tools to report many errors at
once.

Do the same in our loader package. On the upside, this removes
ValidateError, removes a few error returns, and consolidates the API
with go/packages.

The only disadvantage is that we now need to port PrintErrors from
go/packages, which also requires Visit. But it's likely we were going to
need Visit sooner or later, anyway.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -51,6 +51,9 @@ func Run(dir string, args ...string) error {
 	if len(pkgs) == 0 {
 		return fmt.Errorf("no Gunk packages to generate")
 	}
+	if loader.PrintErrors(pkgs) > 0 {
+		return fmt.Errorf("encountered package loading errors")
+	}
 
 	// Record the loaded packages in gunkPkgs.
 	g.recordPkgs(pkgs...)
@@ -91,6 +94,7 @@ func Run(dir string, args ...string) error {
 //
 // Currently, we only generate a FileDescriptorSet for one Gunk package.
 func FileDescriptorSet(dir string, args ...string) (*desc.FileDescriptorSet, error) {
+	// TODO: share code with Run; much of this function is identical.
 	g := &Generator{
 		Loader: loader.Loader{
 			Dir:   dir,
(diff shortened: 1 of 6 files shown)


generate/generate.go
```diff
@@ -51,6 +51,9 @@ func Run(dir string, args ...string) error {
 	if len(pkgs) == 0 {
 		return fmt.Errorf("no Gunk packages to generate")
 	}
+	if loader.PrintErrors(pkgs) > 0 {
```
vishen: We may want to put this in the `FileDescriptorSet` as well. We should probably try and consolidate these two functions if possible.

loader/loader.go
```diff
@@ -201,19 +178,23 @@ func (l *Loader) Load(patterns ...string) ([]*GunkPackage, error) {
...
-func validatePackage(gunkPackage *GunkPackage) *ValidateError {
-	// TODO(vishen): include file location information.
-	errors := []error{}
-	for _, file := range gunkPackage.GunkSyntax {
+const (
+	UnknownError = packages.UnknownError
+	ListError    = packages.ListError
+	ParseError   = packages.ParseError
+	TypeError    = packages.TypeError
+	// Our kins of errors. Add a gap of 10 to be sure we won't conflict with
```
vishen: `s/kin/kinds/`?

generate/generate.go
mvdan: True! Will take a look.

loader/loader.go
mvdan: nice spotting :)

generate/generate.go
mvdan: You're completely right - I had missed this. And I agree we should join the code.

mvdan: Fixed this, added a TODO, and added a regression test for `gunk dump` to check load errors.

Issue #177, opened 2019-01-20
CI: remove the extra steps to download protoc


mvdan:
gunk now dowlnoads protoc as needed, so we shouldn't reimplement that in the CI script.

However, we should make sure that CI doesn't end up downloading protoc a dozen times. Perhaps this means adding a command like `gunk download`, which will download any missing dependencies. Then CI could just run that before running `go test ./...`.

Another possible approach would be for all the test scripts to share the directory where dependencies are downloaded to. But this would require that the download code handles concurrency with other processes, such as grabbing a lock file on disk. I presume we haven't implemented this.

@vishen any ideas?

Issue #175, opened 2019-01-20
generate: dependency downloading needs tests


mvdan:
This code is tricky to get right and easy to break - we badly need regression tests.

The tests should be skipped in short mode and when there's no network. Testscript has support for all that; see `[net]` in https://godoc.org/github.com/rogpeppe/go-internal/testscript.

The tests should also run even if the developer has protoc installed. Otherwise, we might break the protoc download and not notice for a while.

/cc @vishen

vishen:
Yup, I agree. The downloading of protoc was done pretty hastily, so I don't mind if it gets rewritten along with tests.

PR #190, opened 2019-01-28
download: gunk download command

vishen:
Added 'gunk download' command which will download any required, but
missing, tools or packages. Currently, this will only download protoc.

Added test for downloading protoc.

Changed CI to download protoc using 'gunk download'.

Fixes: #177
Fixes: #175

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,20 +9,8 @@ go:
 env:
   - GO111MODULE=on
 
-addons:
-  apt:
-    packages:
-      - unzip
-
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -14,15 +14,8 @@ addons:
     packages:
       - unzip
```
mvdan: Can we remove this now?

vishen: Ah oops, done!

PR #191, opened 2019-01-28
fix app.Terminate to properly terminate kingpin

mvdan:
For better or worse, kingpin's Terminate function needs to actually exit
kingpin's control flow. Since we can't use os.Exit because of
testscript, our only other option is a custom panic type.

One symptom of that bug was that 'gunk -h' displayed the help text twice
- once for the help flag, and a second time since no subcommand was
given.

Fix that, and add a regression test.


mvdan:
Eugh, GitHub formatted my dash as markdown.

PR #192, opened 2019-01-29
loader: support loading lone Gunk files

### LINKED ISSUE #156: loader: support loading gunk files as a single package
Like `go` commands.

```
$ echo 'package foo' >foo.gunk
$ gunk format foo.gunk
error: -: cannot find package "foo.gunk" in any of:
        /home/mvdan/tip/src/foo.gunk (from $GOROOT)
        /home/mvdan/go/land/src/foo.gunk (from $GOPATH)
        /home/mvdan/go/src/foo.gunk
```

This fails because `go/packages` sees the missing `.go` extension, and assumes the argument is a package path, like `foo.com`.

mvdan:
This required a bit of a refactor, as we need to fill the GunkFiles
field early on, and depending on whether we're given Gunk files or not.

While at it, improve the docs, as this logic is getting a bit complex.

Fixes #156.


mvdan:
Note that the diff is large, but much of the changes are moving code around and whitespace.

testdata/scripts/loader_files.txt
```diff
@@ -0,0 +1,15 @@
...
+-- go.mod --
+module testdata.tld/util
+-- p1/p1.gunk --
+package p1  // extra spaces
+-- p1/p1.gunk.golden --
+package p1 // extra spaces
+-- p2/p2.gunk --
+package p2  // extra spaces
+-- p2/p2.gunk.golden --
+package p2 // extra spaces
```
vishen: Should this not have been formatted because it wasn't passed to `gunk format`?

testdata/scripts/loader_files.txt
```diff
@@ -0,0 +1,15 @@
+env HOME=$WORK/home
+
+gunk format p1/p1.gunk
```
vishen: Do we also want to compare the .gunk files to the .gunk.golden files?

loader/loader.go
```diff
@@ -388,8 +414,6 @@ func (l *Loader) toGunkPackage(lpkg *packages.Package) *GunkPackage {
 			}
 		}
 	}
-
-	return pkg
```
vishen: Is this now an implicit return in Go if it gets to the end of the function and their is no return?

testdata/scripts/loader_files.txt
mvdan: Agh, I wrote the first version of this test, and forgot to actually finish it. It's missing at least a handful of lines.

loader/loader.go
mvdan: The old func was `func(*packages.Package) *GunkPackage`, and it's now `func(*GunkPackage)`. So it just modifies the package in-place. The returns now simply skip extra work, and don't actually return anything.

vishen: Ahh, not sure how I didn't see that. Sorry.

mvdan: It's confusing because the context shows:

```
@@ -388,8 +414,6 @@ func (l *Loader) toGunkPackage(lpkg *packages.Package) *GunkPackage {
```

...which isn't useful at all, because that is the *removed* line.

mvdan:
@vishen PTAL

Issue #146, opened 2019-01-10
convert: unable to parse proto files with "." in package name


vishen:
The following will cause errors with the proto parser we are using:

```
// example.proto
syntax = "proto3";
package pokemon.example;
```

results in 

```
go run main.go convert example.proto  --overwrite
error: 1:16: expected ';', found '.'
exit status 1
```

but the following works fine:

```
// example.proto
syntax = "proto3";
package example;
```

I have raised an issue about this - https://github.com/emicklei/proto/issues/102

vishen:
This is actually not an issue with the proto parser, but the way we handle the Gunk (Go) package name, or more accurately, the fact that we don't do anything to to the proto package when when we try and convert it to Gunk package name.

```
// example.proto
syntax = "proto3";
package pokemon.example;
```

converts to

```
// example.gunk
package pokemon.example
```

and since we can't have periods in a package name, we receive an error when trying to format the Gunk code.

I am not too sure how to handle this situation, I can think of a few possibilities:

1) Replace all unexpected characters (like period) with underscores?
2) Ask the user what the package name should be during the convert?
3) Detect what folder the file is in and use that as the package name (although that doesn't always make sense)

I am kind of leaning towards asking the user, do you have any thoughts @mvdan @kenshaw

vishen:
ping @mvdan @kenshaw

vishen:
@mvdan your suggestion seems reasonable to me. 

We may be able to have a cli option that will force set the package, or one that will ask the user, or the default can be your suggestion.

> Perhaps we should also ensure that all output dirs are always under a .gunkconfig - otherwise we could generate broken gunk packages.
I am not quite sure why this would happen? It makes sense to only create the path to proto if the starting directory has a `.gunkconfig` though.

mvdan:
> We may be able to have a cli option that will force set the package, or one that will ask the user, or the default can be your suggestion.

Forcing a package path or using the default algorithm sounds fine. I'm not sure if asking the user would be a good idea; we'd start a precedent for `gunk` to be an interactive tool, and I don't think we want it to be.

> I am not quite sure why this would happen?

If a module has nested gunkconfigs, and the algorithm somehow decided to place converted gunk files in a path within the module that's not under a gunkconfig. We can ignore this or just make it error; I don't think it would happen much in practice.

PR #199, opened 2019-02-01
convert: error if package name isn't Go compatible

vishen:
If the package name of the proto file being converted isn't a Go
compatible package name, e.g: contains a ".".

Updates #146

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 	"text/scanner"
+	"unicode"
 
 	"github.com/emicklei/proto"
 	"github.com/knq/snaker"
@@ -116,6 +117,12 @@ func convertFile(path string, overwrite bool) error {
 		}
 	}
 
+	// Validate that the package name is a a valid
+	// Go package name.
+	if err := b.validatePackageName(); err != nil {
+		return err
+	}
+
 	// Convert the proto package and imports to gunk.
 	translatedPkg, err := b.handlePackage()
 	if err != nil {
@@ -139,9 +146,6 @@ func convertFile(path string, overwrite bool) error {
 		w.WriteString(l)
 		w.WriteString("\n")
 	}
-
-	// TODO: We should run this through the Gunk generator to
-	// make sure that it compiles?
 
 	result := []byte(w.String())
 	result, err = format.Source(result)
@@ -217,6 +221,9 @@ func (b *builder) goType(fieldType string) string {
 	case "uint64", "fixed64":
 		return "uint64"
 	default:
+		// TODO: We return the proto package name unaltered. This
+		// causes issues when a package name is imported or contains
+		// "." or other invalid characters for a package name.
 		// This is either an unrecognised type, or a custom type.
 		return fieldType
 	}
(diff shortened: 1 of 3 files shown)


convert/convert.go
```diff
@@ -611,3 +617,18 @@ func (b *builder) handleLiteralString(lit proto.Literal) (string, error) {
...
+		switch {
+		case c >= '0' && c <= '9':
+		case c >= 'a' && c <= 'z':
+		case c >= 'A' && c <= 'Z':
+		case c == '_':
+		default:
+			return fmt.Errorf("invalid character %q in package name %q", c, pkgName)
+		}
+	}
+	return nil
```
mvdan: this is slightly incorrect :) package names are just identifiers: https://golang.org/ref/spec#Package_clause

So, for example, `"123"` and `""` are invalid, but `"àé"` is valid. I realise that package names in theory shouldn't be empty or just numbers to begin with, but this still makes me uneasy.

Unlucky for us, there's no API in `go/ast` or `astutil` to easily check this. For now, I'd just add this func to the loader package:

```
func IsIdentifier(str string) bool {
	if len(str) == 0 {
		return false
	}
	for i, c := range str {
		if unicode.IsLetter(c) || c == '_' || (i > 0 && unicode.IsDigit(c)) {
			continue
		}
		return false
	}
	return true
}
```

vishen: > I  realise that package names in theory shouldn't be empty or just numbers to begin with, but this still makes me uneasy.

The proto language doesn't allow empty package names, or package names starting with a number, so the proto parser we are using will error on those scenarios (which we have been relying on previously as well).

I wanted to keep the position information of the error just to be useful, and since this won't be used anywhere else yet I will keep this function in this package; if / when we need to use this somewhere else we can look at moving this to the loader or wherever seems appropriate.

vishen: Actually, ignore me, I removed the positional information of the error and didn't realise it. I will add this to the loader.

vishen: I ended up keeping a slightly modified version of your code in the convert package. I decided to keep the character that was causing issues just incase people don't know exactly what a Go identifier is.

mvdan: SGTM! But did you forget to use `loader.IsIdentifier`? I think it should be fine to quote the entire string in the error, like:

> invalid package name %q: identifiers may only contain letters, digits, and underscores

vishen: Ah nope, I decided to leave the identifier check in this function (just to keep information about which character was incorrect).

mvdan: Ok, but then I don't see the point of adding the other func :) Merging for now, anyway.

vishen: I don't know wha other func yo are referring to here, this one is the only one I added :)

mvdan: https://github.com/gunk/gunk/pull/199/files#diff-b38e43f0fb016af3e8cecd2d12d779ecR705 :)

vishen: Oh, that is most definitely not supposed to be there, my apologies :(

mvdan: No worries! Want to send a follow-up PR?

PR #200, opened 2019-02-04
loader: positional information to error messages

### LINKED ISSUE #169: all: ensure errors from gunk files contain positions
As pointed out by @kenshaw:

```
$ gunk generate ./...
error: unsupported field type: []byte
```

This is because we've been lazy and just return errors like `fmt.Errorf("unsupported field type: %v", ftype)`. We should go over them and add positions.

We might be able to keep the change simple by adding positions at a higher level. For example, using the start position of each field for all the errors that originate from converting that field to protobuf.

vishen:
Add positional information to error messages produced by loader.
Moved 'validatePackage' to 'Loader' so the FileSet is populated and
accessible.

Closes #169

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,14 +8,13 @@ require (
 	github.com/ghodss/yaml v1.0.0 // indirect
 	github.com/golang/protobuf v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway v1.6.4
-	github.com/gunk/opt v0.0.0-20190118082020-1439f71a8644 // indirect
 	github.com/knq/ini v0.0.0-20181118015158-a301e724bd35
 	github.com/knq/snaker v0.0.0-20181215144011-2bc8a4db4687
 	github.com/rogpeppe/go-internal v1.1.0
 	github.com/stretchr/testify v1.3.0 // indirect
 	golang.org/x/net v0.0.0-20181106065722-10aee1819953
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -8,14 +8,13 @@ require (
...
 	github.com/grpc-ecosystem/grpc-gateway v1.6.4
-	github.com/gunk/opt v0.0.0-20190118082020-1439f71a8644 // indirect
 	github.com/knq/ini v0.0.0-20181118015158-a301e724bd35
 	github.com/knq/snaker v0.0.0-20181215144011-2bc8a4db4687
 	github.com/rogpeppe/go-internal v1.1.0
 	github.com/stretchr/testify v1.3.0 // indirect
 	golang.org/x/net v0.0.0-20181106065722-10aee1819953
 	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4 // indirect
-	golang.org/x/tools v0.0.0-20190118193359-16909d206f00
+	golang.org/x/tools v0.0.0-20190202235157-7414d4c1f71c
```
mvdan: just curious; any reason why this is being bumped here?

vishen: I didn't intentionally bump this, I ran `go mod tidy` because I had a large number of extra dependencies that I didn't pull in, and this was what was left. Happy to change this back though?

PR #202, opened 2019-02-05
loader: move proto conversion code here

mvdan:
We'll soon need it to start loading proto files as part of Gunk packages.

We can't keep the code inside the convert package, as the convert package
imports loader. To not add a dependency cycle, keep the core of the
converter in the loader package.

Handling files and formatting doesn't really belong in the loader, so leave
those out. This process does make the convert package quite small, but
that's okay.


mvdan:
DO NOT MERGE until #201 is merged.

mvdan:
rebased on master; ok to merge.

PR #206, opened 2019-04-18
convert: handle duplicate message

cedric-parisi:
- Ensure that there is no duplicate between nested and top-level message. 
- Rename nested message to `Parent_Child` instead of `Child` when creating the top level struct

### DIFF
--- a/loader/proto.go
+++ b/loader/proto.go
@@ -26,8 +26,9 @@ func ConvertFromProto(w io.Writer, r io.Reader, filename string) error {
 
 	// Start converting the proto declarations to gunk.
 	b := builder{
-		filename:    filename,
-		importsUsed: map[string]string{},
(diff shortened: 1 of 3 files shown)


loader/proto.go
```diff
@@ -88,6 +89,9 @@ type builder struct {
 	// Mostly these will be Gunk annotations. Import name will be
 	// mapped to its possible named import.
 	importsUsed map[string]string
+
+	// Holds existings declaration to avoid duplicate
+	existingDeclarations map[string]bool
```
mvdan: you could call this `existingDecls`, as that's a common abbreviation for the word.

cedric-parisi: Indeed. I've renamed to `existingDecls` and amended my commit.

PR #210, opened 2019-04-22
option: add openapiv2 support

cedric-parisi:
This PR adds OpenAPI v2 support:
- convert from proto to gunk by adding `openapiv2.Schema` and `openapiv2.Operation` to `+gunk` annotation
- generate from gunk to all.swagger.json 

minor fix on enum names by adding the enum type name to its value in case of duplicate


generate/generate.go
```diff
@@ -762,6 +788,29 @@ func (g *Generator) methodOptions(method *ast.Field) (*desc.MethodOptions, error
 				return nil, err
 			}
 			g.addProtoDep("google/api/annotations.proto")
+		case "github.com/cedric-parisi/opt/openapiv2.Operation":
```
mvdan: surely this should be `gunk/opt` now?

loader/proto.go
```diff
@@ -353,6 +422,13 @@ func (b *builder) handleEnum(e *proto.Enum) error {
 			continue
 		}
 
+		// Check if there is already an existing enum field with this name
+		if _, ok := b.existingDecls[ef.Name]; ok {
```
mvdan: simpler - `if b.existincDecls[ef.Name] {`

loader/proto.go
```diff
@@ -429,10 +550,14 @@ func (b *builder) handleService(s *proto.Service) error {
 						}
 					default:
 						method = n
-						url, err = b.handleLiteralString(*l.Literal)
+						tmp, err := b.handleLiteralString(*l.Literal)
 						if err != nil {
 							return b.formatError(opt.Position, "option for %q should be a string (url)", method)
 						}
+						var re = regexp.MustCompile(`\{(.*?)\}`)
```
mvdan: compile this regex as a global, so you don't have to keep recomipiling it. also, prefer `x := value` over `var x = value` for local variables.

generate/generate.go
cedric-parisi: indeed I missed it before submit...

.gitignore
```diff
@@ -2,3 +2,5 @@
 /google
 /.bin
 /build/
+/pb/
+.gunkconfig
```
mvdan: don't add this, because we do have some gunk configs in examples and testdata

PR #211, opened 2019-04-24
loader: add convert comment

### LINKED ISSUE #197: convert: comments not capitalised
When gunk converts from proto to Gunk, we force all variables and fields to capitalised `CamelCase`, but we don't update the comment at the moment. 

```
# example.proto
message Error {
    // error_code is an error code
    string error_code = 1; 
}
```

```
# example.gunk
type Error struct {
    // error_code is an error code
    ErrorCode string `pb:"1"`
}
```

Gunk should update the comment to contain the updated field if the name is at the start of the comment.

cuonglm:
Fixes #197

### DIFF
--- a/loader/proto.go
+++ b/loader/proto.go
@@ -222,6 +222,9 @@ func (b *builder) handleMessageField(w *strings.Builder, field proto.Visitee) er
 		return fmt.Errorf("unhandled message field type %T", field)
 	}
 
+	if comment != nil && strings.HasPrefix(strings.TrimSpace(comment.Message()), name) {
+		comment.Lines[0] = strings.Replace(comment.Message(), name, snaker.ForceCamelIdentifier(name), 1)
+	}
 	if repeated {
 		typ = "[]" + typ
 	}
--- a/testdata/scripts/convert_comment.txt
+++ b/testdata/scripts/convert_comment.txt
@@ -1,0 +1,24 @@
+env HOME=$WORK/home
+
+gunk convert util.proto
+cmp util.gunk util.gunk.golden
+
+-- util.proto --
+syntax = "proto3";
+
+package util;
+
+message Error {
+	// error_code is an error code,
+	// the second line should not be changed
+	string error_code = 1;
+}
+
+-- util.gunk.golden --
+package util
+
+type Error struct {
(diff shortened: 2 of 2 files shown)


testdata/scripts/convert_comment.txt
```diff
@@ -0,0 +1,22 @@
+env HOME=$WORK/home
+
+gunk convert util.proto
+cmp util.gunk util.gunk.golden
+
+-- util.proto --
```
mvdan: this test script should also test multi-line doc comments.

loader/proto.go
```diff
@@ -639,3 +641,10 @@ func (b *builder) validatePackageName() error {
 	}
 	return nil
 }
+
+// updateComment updates first line comment if it start with name.
+func (b *builder) updateComment(comment *proto.Comment, name string) {
+	if comment != nil && strings.HasPrefix(strings.TrimSpace(comment.Message()), name) {
+		comment.Lines[0] = strings.Replace(comment.Message(), name, snaker.ForceCamelIdentifier(name), 1)
+	}
```
mvdan: invert the condition and early return.

loader/proto.go
```diff
@@ -207,12 +207,14 @@ func (b *builder) handleMessageField(w *strings.Builder, field proto.Visitee) er
...
 		typ = b.goType(field.Type)
 		sequence = field.Sequence
+		b.updateComment(field.Comment, name)
 		comment = field.Comment
 		repeated = field.Repeated
 		options = field.Options
 	case *proto.MapField:
 		name = field.Field.Name
 		sequence = field.Field.Sequence
+		b.updateComment(field.Comment, name)
```
mvdan: can you not move this line to right after the switch case, and deduplicate it?

in fact, that would probably mean you no longer need the method.

testdata/scripts/convert_comment.txt
cuonglm: it seems that we currently doesn't handle the form `/* ... */`. We always transform to `//` form. Should I raise new issue?

mvdan: No, I don't think that's necessary.

PR #212, opened 2019-04-29
stream: handle streaming for convert and generate

### LINKED ISSUE #133: Support for grpc streaming
Should we add support for grpc streaming? 

If we did, we could have it as a Gunk annotation `+gunk.GRPCStream` or something, or apart of the syntax `Echo(chan Message) chan Message`?

Not entirely sure how protoc generates the `FileDescriptorProto` with `stream` set for a service, need to investigate further.

cedric-parisi:
Add stream support for both convert and generate commands.

fix #133


generate/generate.go
```diff
@@ -905,25 +905,31 @@ func (g *Generator) convertMap(parentName, fieldName string, mapTyp *types.Map)
...
+		return nil, nil, fmt.Errorf("unsupported parameter type: %v", param)
 	}
 	if label == desc.FieldDescriptorProto_LABEL_REPEATED {
-		return nil, fmt.Errorf("parameter type should not be repeated")
+		return nil, nil, fmt.Errorf("parameter type should not be repeated")
 	}
-	return &tname, nil
+
+	isStream := proto.Bool(false)
+	if strings.Contains(param.String(), "chan") {
```
mvdan: don't stringify a type to see if it's a channel; this will break, for example, if one does `type chanFoo int`. instead, check the "type of the type" agains `types.Chan`.

PR #213, opened 2019-05-01
convert: handle import proto files

### LINKED ISSUE #198: convert: handle imports
When converting a proto file to Gunk, imports are largely ignored, and types from those imports are outputted  as is:

```
# other/file.proto
package other.file;

message Msg {
    string msg = 1;
}
```

```
# example.proto
package example;

import "other/file.proto"

message Request {
    other.file.Msg msg = 1;
}
```

converts to

```
package example

import (
    // other/file.proto
)

type Request struct {
    Msg other.file.Msg `pb:"1"`
}
```

This causes issues for a couple of reasons. 

1) Imports are never actually imported
2) Package names of these imports aren't normalised to be compatible as Go package names.

We could load (and convert) any imports we find for the file being converted, but we won't necessarily know where to find the file unless `gunk convert` is run from the same place as `protoc` is usually run?

We could attempt to force the user to add the file when running `gunk convert`?

cedric-parisi:
Add `[convert]` section to handle import when converting `.proto` to `.gunk`:
If `import_path` is provided within that section, gunk will pass this path to protoc in `LoadProto` method.
proto files passed will be parsed to retrieve `go_package` option value and added as import in resulting gunk file

Fix #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -478,6 +478,31 @@ $ gunk convert /path/to/file.proto
 $ gunk convert /path/to/protobuf/directory
 ```
 
+If your `.proto` is referencing another `.proto` from another directory, 
+you can add `import_path` in the global section of your `.gunkconfig`.
+If you don't provide `import_path` it will only search in the root directory.
+
+
+```ini
+import_path=relative/path/to/protobuf/directory
+```
+
+> The path to provide is relative from the `.gunkconfig` location.
+
+Furthermore, the referenced files must contain:
+
+```proto
+option go_package="path/of/go/package";
+```
+
+The resulting `.gunk` file will contain the import path as defined in `go_package`:
+
+```go
+import (
+	name "path/of/go/package"
+)
+```
+
 ## About
 
 Gunk is developed by the team at [Brankas][brankas], and was designed to
--- a/config/config.go
+++ b/config/config.go
@@ -75,6 +75,7 @@ func (g Generator) OutPath(packageDir string) string {
 type Config struct {
 	Dir        string
 	Out        string
+	ImportPath string
 	Generators []Generator
 }
 
@@ -115,6 +116,7 @@ func Load(dir string) (*Config, error) {
 					cfg.Generators[i].Out = cfg.Out
 				}
 			}
+
 			cfgs = append(cfgs, cfg)
 		}
 
@@ -211,7 +213,9 @@ func load(reader io.Reader) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
-		config.Generators = append(config.Generators, *gen)
+		if gen != nil {
+			config.Generators = append(config.Generators, *gen)
+		}
 	}
 	return config, nil
 }
@@ -249,6 +253,8 @@ func handleGlobal(config *Config, section *parser.Section) error {
 		switch k {
 		case "out":
 			config.Out = v
+		case "import_path":
+			config.ImportPath = v
 		default:
 			return fmt.Errorf("unexpected key %q in global section", k)
 		}
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -7,31 +7,48 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/gunk/gunk/config"
 	"github.com/gunk/gunk/format"
+	"github.com/gunk/gunk/generate"
 	"github.com/gunk/gunk/loader"
 )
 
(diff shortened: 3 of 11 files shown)


mvdan:
Please don't include information in the PR body and not in the commit message; only the latter survives in git.

Also, it would be good if the commit message explained why a flag is necessary. Without having looked at the code carefully, it's not obvious to me that we need a flag :)

convert/convert.go
```diff
@@ -28,10 +29,22 @@ func run(path string, overwrite bool) error {
 	if err != nil {
 		return err
 	}
+
+	// Look for a .gunkconfig
+	absPath, _ := filepath.Abs(path)
+	cfg, err := config.Load(filepath.Dir(absPath))
```
mvdan: I think this would be simpler if you just did `Load("")`, which will use $PWD. We do use gunk packages as the base for finding gunkconfig files, but I think that makes less sense when we're dealing with single protobuf files.

convert/convert.go
```diff
@@ -28,10 +29,22 @@ func run(path string, overwrite bool) error {
...
+	if err != nil && err != config.ErrNoConfig {
+		return fmt.Errorf("unable to load .gunkconfig: %v", err)
+	}
+	var convertor *config.Convertor
+	if cfg != nil && cfg.Convertor != nil {
+		convertor = cfg.Convertor
+	}
```
mvdan: I think this can be much simpler:

```
var t T
cfg, err := config.Load("")
if err == nil {
    t = cfg.T
}
```

I don't think differentiating config load errors is necessary. If `convert` doesn't need a config file, I don't see why it should error if an invalid config file is on disk.

config/config.go
```diff
@@ -31,6 +34,12 @@ type Generator struct {
 	Out       string
 }
 
+type Convertor struct {
+	Params     []KeyValue
```
mvdan: these parameters aren't used?

config/config.go
```diff
@@ -31,6 +34,12 @@ type Generator struct {
...
+	ImportPath string
```
mvdan: I was actually imagining this to be a list. Imagine that we want to find proto files in multiple places, like:

```
import_path=our/proto/files
import_path=vendored/proto/files
```

This doesn't have to work today, but I think limiting it to zero or one directories is a bit weird.

README.md
```diff
@@ -478,6 +478,28 @@ $ gunk convert /path/to/file.proto
 $ gunk convert /path/to/protobuf/directory
 ```
 
+If your `.proto` is referencing another `.proto`, you must add a `[convert]` section in your `.gunkconfig`. If you don't provide `import_path` it will only search in the directory gunk is executed.
+
+
+```ini
+[convert]
+import_path= /path/to/protobuf/directory
```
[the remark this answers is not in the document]
cedric-parisi: yes, I'll remove [convert] section and just use a global import path

convert/convert.go
cedric-parisi: Doing `Load("")` will work only if you are executing gunk from the same root directory than your gunkconfig. 
For instance, I'm doing `go build && ./gunk convert ../../examples/pb/util.proto --overwrite`. Like that I'm covering both cases.

config/config.go
cedric-parisi: It seems that currently the dep used to parse `ini` files doesn't provide a way to have the same key multiple times. It will always return `our/proto/files`

mvdan: We've agreed to look into this in a future PR.

convert/convert.go
mvdan: Hmm, `Load("")` simply looks for the config starting at the current directory, and then walking up the chain of parent directories until it finds the root of the project (a gunk config, or a git directory, etc), or the root of the filesystem. Isn't that enough?

README.md
```diff
@@ -478,6 +478,29 @@ $ gunk convert /path/to/file.proto
 $ gunk convert /path/to/protobuf/directory
 ```
 
+If your `.proto` is referencing another `.proto` from another directory, 
+you can add `import_path` in the global section of your `.gunkconfig`.
+If you don't provide `import_path` it will only search in the root directory.
+
+
+```ini
+import_path=relative/path/to/protobuf/directory
```
mvdan: clarify that this is relative to the gunk project root, i.e. where the config is

mvdan:
CI is red too?

PR #216, opened 2019-05-24
convert: add swagger object

cedric-parisi:
- Handle complete swagger definition for convert command

### DIFF
--- a/loader/proto.go
+++ b/loader/proto.go
@@ -5,17 +5,17 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"text/scanner"
 	"unicode"
 
-	"github.com/golang/protobuf/ptypes/any"
-
-	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
-
 	"github.com/emicklei/proto"
 	"github.com/knq/snaker"
+
+	"github.com/gunk/opt/openapiv2"
 )
 
 var urlVarRegexp = regexp.MustCompile(`\{(.*?)\}`)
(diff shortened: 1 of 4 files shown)


loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+		name := s.Field(i).Name
+		value := v.FieldByName(name)
+		if !isNilOrEmpty(value.Interface()) {
+			k := value.Kind()
+			switch {
+			case k >= reflect.Int && k <= reflect.Int64,
+				k == reflect.Float32 || k == reflect.Float64,
+				k >= reflect.Uint && k <= reflect.Uint64,
+				k == reflect.String,
+				k == reflect.Bool:
```
mvdan: I'm not sure I get this logic. Is this to check if the kind is a builtin type?

loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+				// The only way to know if the value
+				// is of type alias is to check its type.
+				if isGoType(value.Type().String()) {
+					b.format(w, 0, nil, "// %s: %v,\n", name, value)
+				} else {
+					pkg := strings.Split(reflect.TypeOf(value.Interface()).String(), ".")[0]
+					b.format(w, 0, nil, "// %s: %s.%v,\n", name, pkg, value)
+				}
+			case k == reflect.Map:
```
mvdan: all of these are based on `k`, so perhaps do a `switch k`? I don't imagine the first case would get more complex, as there aren't that many kinds.

loader/proto.go
mvdan: Another option would be to handle map/ptr/slice separately, reject/ignore specific types like array, and then make this the "default" case.

loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+		case "responses":
+			key, resp, err := b.convertResponse(v)
+			if err != nil {
+				return nil, err
+			}
+			swagger.Responses[key] = resp
+		case "swagger":
+			swagger.Swagger = v.SourceRepresentation()
+		case "base_path":
+			swagger.BasePath = v.SourceRepresentation()
```
mvdan: I can't help but find this too repetitive. Any thoughts on automating this more?

testdata/scripts/convert_complete_swagger.txt
```diff
@@ -0,0 +1,305 @@
...
+      summary : "Retrieves message"
+      tags : "Message"
+    };
+    option (google.api.http) = {
+      get : "/v1/message/{Name}"
+    };
+  }
+}
+
+-- util.gunk.golden --
```
mvdan: unused? or did you forget to do the cmp with the two files?

loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+				b.format(w, 0, nil, "// },\n")
+			case k == reflect.Slice:
+				b.format(w, 0, nil, "// %s: %s{\n", name, reflect.TypeOf(value.Interface()))
+				for i := 0; i < value.Len(); i++ {
+					val := value.Index(i).Interface()
+					switch reflect.TypeOf(val).Kind() {
+					case reflect.Ptr:
+						// TODO (ced)
+					default:
+						if isGoType(reflect.TypeOf(val).String()) {
```
mvdan: This seems a bit weird. When would a `TypeOf` not be a go type?

Also, whatever you want to know about the type, you can do so without string handling. For example, to know if a type is named versus a simple Go type, just use the `Name() string` method.

loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+				k == reflect.Float32 || k == reflect.Float64,
+				k >= reflect.Uint && k <= reflect.Uint64,
+				k == reflect.String,
+				k == reflect.Bool:
+				// The only way to know if the value
+				// is of type alias is to check its type.
+				if isGoType(value.Type().String()) {
+					b.format(w, 0, nil, "// %s: %v,\n", name, value)
+				} else {
+					pkg := strings.Split(reflect.TypeOf(value.Interface()).String(), ".")[0]
```
mvdan: Obtain the pkg via the PkgPath method on Type.

loader/proto.go
```diff
@@ -751,6 +723,454 @@ func (b *builder) handlePackage() (string, error) {
...
+				for i := 0; i < value.Len(); i++ {
+					val := value.Index(i).Interface()
+					switch reflect.TypeOf(val).Kind() {
+					case reflect.Ptr:
+						// TODO (ced)
+					default:
+						if isGoType(reflect.TypeOf(val).String()) {
+							b.format(w, 0, nil, "// %v,\n", val)
+						} else {
+							pkg := strings.Split(reflect.TypeOf(val).String(), ".")[0]
```
mvdan: PkgPath, like before

loader/proto.go
cedric-parisi: When you have an enum, the `Kind()` will be the underlying type of the alias and the  `TypeOf()` is the alias. In the gunk annotation I want to display the enum name and not its underlying value: 
```
Type: openapiv2.TYPE_API_KEY ,
```
for instance and not 
```
Type: 1,
```

testdata/scripts/convert_complete_swagger.txt
cedric-parisi: Unused, I'll remove it, I wanted to do a cmp but each run returns a different order for the map or slices.

loader/proto.go
cedric-parisi: Indeed, I'll do it using reflection

testdata/scripts/convert_complete_swagger.txt
mvdan: Hmm, the output should be deterministic, ideally... otherwise one can't re-run gunk convert, as it would change files in git.

mvdan: Of course we can look at that as part of a separate PR, but it will also make your tests simpler.

loader/proto.go
mvdan: If you want the name of a named type, you can use the Name method - otherwise Kind can tell you what kind the type is. String is too sloppy for this stuff, IMO. But we can clean this up after this PR if you're not sure what I mean.

testdata/scripts/convert_complete_swagger.txt
cedric-parisi: Yes that would be the best scenario, I'll get into it into another PR if that's ok

PR #219, opened 2019-05-27
generate: generate complete swagger definition from gun annotations

cedric-parisi:
- Add complete swagger definition for generate command
- Move `setValue` to a shared pkg as this function is also used by `generate`
- Update tests with latest version of `gunk/opt`


utils/setter.go
```diff
@@ -0,0 +1,43 @@
+package utils
```
mvdan: perhaps a less generic name, like `reflectutil`?

utils/setter.go
```diff
@@ -0,0 +1,43 @@
...
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+// SetValue assign value to the field
+// name of the struct dest.
+// Returns an error if name is not found or dest cannot
+// be addressed.
+func SetValue(name string, dest interface{}, value string) error {
```
mvdan: I'd prefer if the signature was clearer, like:

```
SetValue(structPtr interface{}, fieldName, value string)
```

Also, I'd move all the strconv.Unquote work into this func.

utils/setter.go
```diff
@@ -0,0 +1,43 @@
...
+	tmp := reflect.Indirect(reflect.ValueOf(dest)).FieldByNameFunc(func(s string) bool {
+		return strings.ToUpper(s) == strings.ToUpper(name)
```
mvdan: Why do you need to do a case insensitive search?

In any case, this could be strings.EqualFold.

utils/setter.go
```diff
@@ -0,0 +1,43 @@
...
+	})
+	if !tmp.IsValid() {
+		return fmt.Errorf("%s was not found in %T", name, dest)
+	}
+	if !tmp.CanAddr() {
+		return fmt.Errorf("%s from %T cannot be addressed", name, dest)
+	}
+	var v interface{}
+	var err error
+	switch k := reflect.TypeOf(tmp.Interface()).Kind(); k {
```
mvdan: please stop doing `reflect.TypeOf(v.Interface())`; just do `v.Type()`

testdata/scripts/convert_options_openapiv2.txt
```diff
@@ -57,7 +57,7 @@ type Account struct {
 	//         JSONSchema: openapiv2.JSONSchema{Title: "Account detail", Description: "Detail information of account"},
 	// }
 	// +gunk openapiv2.Schema{
-	//         Example: map[string]string{"value": `{"success":true,"error_code":"0","error_message":""}`},
+	//         Example: openapiv2.Any{Value: []byte(`{"success":true,"error_code":"0","error_message":""}`)},
```
mvdan: was this on purpose?

generate/generate.go
```diff
@@ -501,6 +510,267 @@ func (g *Generator) fileOptions(pkg *loader.GunkPackage) (*desc.FileOptions, err
...
+			info.License = license
+		}
+	}
+	return info, nil
+}
+
+func (g *Generator) convertJSONSchema(lit *ast.CompositeLit) (*options.JSONSchema, error) {
+	jsonSchema := &options.JSONSchema{}
+	for _, elt := range lit.Elts {
+		name := elt.(*ast.KeyValueExpr).Key.(*ast.Ident).Name
```
mvdan: you can deduplicate all this `elt.(*ast.KeyValueExpr)` work, here and elsewhere, like:

```
name := elt.(*ast.KeyValueExpr).Key.(*ast.Ident).Name
value := elt.(*ast.KeyValueExpr).Value
```

utils/setter.go
mvdan: Also, as a nit, I'd probably just make this func panic instead of returning an error. An error should only occur if there's a bad bug in our code, in which case panicking is fine. And your code will get much cleaner.

cedric-parisi: I can't move the Unquote in this func as it is not needed for the `convert`.

utils/setter.go
cedric-parisi: This is because protoc doesn't follow the initialisms in their namings. So in `options.pb.go`, you'll find `TokenUrl` and in gunk you'll find `TokenURL`.

mvdan: I see - could you add this as a comment in the code?

testdata/scripts/convert_options_openapiv2.txt
cedric-parisi: yes, upgrading to the last version of `gunk/opt` introduced this change in `Example` definition.

utils/setter.go
cedric-parisi: sorry, I didn't know about that before the last PR and I missed this one...

mvdan: no worries :)

PR #222, opened 2019-06-04
reflectutil: move more reflect magic here

mvdan:
We can move a lot of work to this helper function. For example, grabbing
the name from an identifier key, the value from a literal, or unquoting
a value string.

Also make it support converting types, so we can assign an int32 value
to an enum field, since it has a named type.

We can also remove many uses of snaker; the field search was already
case-insensitive, so removing all underscores is all that we need to
make it work.

This means we can drastically simplify and unify many uses of SetValue,
making the meat of the code simpler and less verbose.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -515,26 +515,19 @@ func (g *Generator) convertOperation(lit *ast.CompositeLit) (*options.Operation,
 	for _, elt := range lit.Elts {
 		kv := elt.(*ast.KeyValueExpr)
 		switch name := kv.Key.(*ast.Ident).Name; name {
+		default:
+			reflectutil.SetValue(op, name, kv.Value)
 		case "Tags":
 			for _, tag := range kv.Value.(*ast.CompositeLit).Elts {
 				val, _ := strconv.Unquote(tag.(*ast.BasicLit).Value)
 				op.Tags = append(op.Tags, val)
 			}
-		case "Description",
-			"Summary",
-			"OperationID":
-			val, _ := strconv.Unquote(kv.Value.(*ast.BasicLit).Value)
-			reflectutil.SetValue(op, name, val)
-		case "Deprecated":
-			val := kv.Value.(*ast.Ident).Name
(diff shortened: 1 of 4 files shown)


generate/generate.go
```diff
@@ -720,30 +709,19 @@ func (g *Generator) convertSecurityDefinitions(lit *ast.CompositeLit) (*options.
 			for _, v := range ss.(*ast.KeyValueExpr).Value.(*ast.CompositeLit).Elts {
 				n := v.(*ast.KeyValueExpr)
 				switch key := n.Key.(*ast.Ident).Name; key {
-				case "Type":
-					sel, ok := options.SecurityScheme_Type_value[n.Value.(*ast.SelectorExpr).Sel.Name]
+				case "Type", "In", "Flow":
+					enumMap := proto.EnumValueMap(prefix + key)
```
mvdan: I experimented with a way to not hard-code the `Foo_Type_value` maps, and it kinda works, but requires me to have the prefix hard-coded. Not sure if there's a cleaner way to do it.

reflectutil/setter.go
```diff
@@ -2,51 +2,104 @@ package reflectutil
...
+	}
+
 	// Performs a case insensitive search because generated structs by
 	// protoc don't follow the best practices from the go naming convention:
-	// initialisms should be all capitals
-	tmp := reflect.Indirect(reflect.ValueOf(structPtr)).FieldByNameFunc(func(s string) bool {
-		return strings.EqualFold(s, name)
+	// initialisms should be all capitals.
+	// Remove underscores too, so that foo_bar matches the field FooBar.
+	nameStr = strings.ReplaceAll(nameStr, "_", "")
```
cedric-parisi: I think this is why travis fails

mvdan: Let me drop 1.11 support ;) after all, 1.13 is around the corner, and 1.11 has somewhat bad module support.

mvdan:
I'm blocked on this, so I'll assume you're fine with it to send another PR :)

PR #224, opened 2019-06-05
reflectutil: support all maps and enums

mvdan:
Gets rid of almost all the rest of the boilerplate code. The only pieces
left are because SetValue can only deal with struct fields.

While at it, unindent and simplify fromStructToAnnotation.

Finally, make the final section of valueFor parse strings alone,
handling the special enum case separately.

### DIFF
--- a/convert/convert.go
+++ b/convert/convert.go
@@ -98,6 +98,9 @@ func convertFile(path string, overwrite bool, importPath string, protocPath stri
 
 	result, err := format.Source(b.Bytes())
 	if err != nil {
+		// Also print the source being formatted, since the go/format
+		// error often points at a specific error in one of its lines.
+		fmt.Fprintln(os.Stderr, b.String())
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


mvdan:
No more boilerplate :tada: 

Still not done, but 90% there.

convert/convert.go
```diff
@@ -98,6 +98,7 @@ func convertFile(path string, overwrite bool, importPath string, protocPath stri
 
 	result, err := format.Source(b.Bytes())
 	if err != nil {
+		fmt.Println(b.String())
```
cedric-parisi: You missed this one I guess :)

mvdan: This was actually on purpose. Many things I tried added syntax errors, so format.Source would say "invalid syntax at line 6 ...".

So I think it's reasonable for convert to output the generated code when erroring like this. Not ideal, but at least better than giving an obscure error with no more info.

cedric-parisi: OK! maybe the gunk/log instead ? Or a comment ? 
I'm afraid that this could be removed at some point because it's `fmt.Println`

mvdan: Done - it's now a Fprintln with a comment.

PR #230, opened 2019-06-25
docgen: init docgen with pot generation

cedric-parisi:
Init docgen plugin with pot generation from the all.proto computed
in generatePlugin from generate cmd.

While at it, fix tests which are failing due to an
out of date grpc-gateway version.


docgen/extract/extract.go
```diff
@@ -0,0 +1,78 @@
...
+// Run extract all string from swagger options and store them into pot builder.
+func Run(file *google_protobuf.FileDescriptorProto) (pot.Builder, error) {
+	pb := pot.NewBuilder()
+
+	if proto.HasExtension(file.GetOptions(), options.E_Openapiv2Swagger) {
+		ext, err := proto.GetExtension(file.GetOptions(), options.E_Openapiv2Swagger)
+		if err != nil {
+			return nil, err
+		}
+		pb.AddTranslations(toTranslate(*ext.(*options.Swagger)))
```
mvdan: toTranslate takes an interface, so the type cast here seems unnecessary

docgen/extract/extract.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	for _, s := range file.GetService() {
+		for _, m := range s.GetMethod() {
+			if proto.HasExtension(m.GetOptions(), options.E_Openapiv2Operation) {
+				ext, err := proto.GetExtension(m.GetOptions(), options.E_Openapiv2Operation)
+				if err != nil {
+					return nil, err
+				}
+				pb.AddTranslations(toTranslate(*ext.(*options.Operation)))
```
mvdan: same as above

docgen/extract/extract.go
```diff
@@ -0,0 +1,78 @@
...
+		if err != nil {
+			return nil, err
+		}
+		pb.AddTranslations(toTranslate(*ext.(*options.Swagger)))
+	}
+
+	for _, s := range file.GetService() {
+		for _, m := range s.GetMethod() {
+			if proto.HasExtension(m.GetOptions(), options.E_Openapiv2Operation) {
+				ext, err := proto.GetExtension(m.GetOptions(), options.E_Openapiv2Operation)
```
mvdan: can't you just do GetExtension, and skip the iteration if the error is "not found"? otherwise these two lines are a bit repetitive.

docgen/pot/pot.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/gunk/gunk/reflectutil"
+)
+
+const (
+	pot = `# SOME DESCRIPTIVE TITLE.
```
mvdan: I'd give this a better name - otherwise it's all very confusing, with a global `pot` in the file `pot.go` in the package `pot`

docgen/docgen.go
```diff
@@ -0,0 +1,54 @@
...
+	pg "github.com/gunk/gunk/plugin"
+)
+
+func main() {
+	if err := pg.RunPlugin(new(plugin)); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type plugin struct{}
```
mvdan: what about calling this `docgen` or `docPlugin`? then you don't have to rename the import

docgen/docgen.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/gunk/gunk/docgen/extract"
+	pg "github.com/gunk/gunk/plugin"
+)
+
+func main() {
+	if err := pg.RunPlugin(new(plugin)); err != nil {
+		log.Fatal(err)
+	}
+}
```
mvdan: nit: you can make the plugin package export a main-like function, such as:

```
func main() { plugin.RunMain(new(plugin)) }
```

RunMain is basically a shared main func. The other added benefit is less stutter; you don't have a `foo.RunFoo` name.

docgen/docgen.go
```diff
@@ -0,0 +1,54 @@
...
+			source = f
+			break
+		}
+	}
+
+	if source == nil {
+		return nil, fmt.Errorf("no file to generate")
+	}
+
+	name := proto.String(filepath.Join(filepath.Dir(source.GetName()), "messages.pot"))
```
mvdan: nit: this would be a bit clearer like:

```
dest := filepath.Join(filepath.Dir(source.GetName()), "messages.pot")
...
Name: proto.string(dest)
```

mvdan: (in particular, "name" doesn't really say much, and I had to read the rest of the code to realise that it was the *desination* path name)

docgen/extract/extract.go
cedric-parisi: Yes. I had to add an indirect in the toTranslate func otherwise `NumField` would panic as ext are ptr to swagger struct.

mvdan: Sorry, I misread that code as `(*ext).(*options.Swagger)` :)

PR #231, opened 2019-07-03
docgen: generate service doc based on markdown template

cedric-parisi:
Refactor docgen pkg to generate service doc and pot files
at the same time.


docgen/docgen.go
```diff
@@ -33,18 +33,23 @@ func (p *docPlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.Co
 		return nil, fmt.Errorf("no file to generate")
 	}
 
-	dest := proto.String(filepath.Join(filepath.Dir(source.GetName()), "messages.pot"))
+	base := filepath.Join(filepath.Dir(source.GetName()))
 
-	e, err := extract.Run(source)
+	var buff bytes.Buffer
```
mvdan: nit: they're usually called `buf`

docgen/generate/generate.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	f := make([]byte, fi.Size())
+	if _, err := file.Read(f); err != nil {
+		if err != io.EOF {
+			return nil, err
+		}
+	}
```
mvdan: Why not just use `ioutil.ReadAll`? I know there's another piece of code that does a manual read, but I think this is wrong.

For example, if you ask to read 500 bytes, the reader can tell you it read 250, and then read 250 the next time. That's valid, but it would break with this code here.

It also means you can get rid of all the unnecessary Stat stuff. You can simplify the other chunk of code too, if that's not too much for this PR :)

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+			if tf == mf.Name {
+				res[i] = mf
+			}
+		}
+	}
+
+	return res, nil
+}
+
+// Get ...
```
mvdan: finish this or remove it

docgen/parser/model.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Comment describes a comment.
+type Comment struct {
+	Leading  string
+	Trailing string
+	Detached []string
+}
+
+// Comments wrap all comments from the proto file.
```
mvdan: `wraps`

mvdan:
Oh, you fixed CI while I was reviewing :)

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+	return res, nil
+}
+
+// Get gets a comment by its location path.
+func (c Comments) Get(path string) *Comment {
+	if val, ok := c[path]; ok {
+		return val
+	}
+	return &Comment{}
+}
```
mvdan: I think it would be simpler to remove this Comments type and the method, and use the map directly with a helper. Like:

```
func nonNilComment(c *Comment) *Comment {
    if c == nil { return &Comment{})
    return c
}

[...]

Comment: nonNilComment(comments["str"]),
```

This is less code, the caller side is more intuitive, and there's less getter magic going on.

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return res
+}
+
+func getType(f *google_protobuf.FieldDescriptorProto) string {
+	if t := f.GetTypeName(); t != "" {
+		s := strings.Split(t, ".")
+		return s[len(s)-1]
+	}
+	return strings.ToLower(strings.Replace(f.GetType().String(), "TYPE_", "", 1))
```
mvdan: You probably want to use `TrimPrefix` instead, which is simpler and less prone to confusing behavior

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/golang/protobuf/proto"
+	google_protobuf "github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/httprule"
+	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
+	method "google.golang.org/genproto/googleapis/api/annotations"
+)
+
+const (
+	messageFlag = 4
+	fieldFlag   = 2
```
mvdan: document where these originate from

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	return res, nil
+}
+
+func parseRequest(rule *method.HttpRule, message *Message) (*Request, error) {
+	var verb, uri string
+
+	switch p := rule.GetPattern().(type) {
+	case *method.HttpRule_Get:
+		verb = "GET"
```
mvdan: weird that this isn't possible via some method or func, like `p.MethodName()` or something.

docgen/parser/parser.go
```diff
@@ -0,0 +1,213 @@
...
+		verb = "DELETE"
+		uri = p.Delete
+	default:
+		return nil, fmt.Errorf("%t not supported", p)
+	}
+
+	var body *Message
+	if rule.GetBody() != "" {
+		body = message
+	}
```
mvdan: sorry, this piece of code is confusing. why do you check `rule.GetBody` but never use it? I think this needs a comment

mvdan:
Please merge once you're happy with the changes, as I'll be away due to the timezone difference.

docgen/parser/parser.go
cedric-parisi: Indeed, GetPattern returns the private interface `isHttpRule_Pattern` which only provides `isHttpRule_Pattern()` and each `HttpRule_<verb>` has one method which is the matching <verb>

PR #238, opened 2019-08-04
docgen: add support for nested messages

mgfcmartinez:
Make docgen support displaying additional information in the generated
markdown file about nested messages in both request and response bodies.
It can now handle arbitrarily nested messages.

Also prevent the Annex section from being printed when it has no
contents.


docgen/parser/parser.go
```diff
@@ -90,9 +92,51 @@ func parseMessages(pkgName string, comments map[string]*Comment, messages []*goo
...
+			name := typ.QualifiedName
+			if !typ.IsMessage || seen[name] {
+				continue
+			}
+
+			seen[name] = true
+
+			nm, ok := messages[name]
+			if !ok { // shouldn't happen in theory, but just to be safe
+				continue
```
mvdan: if it shouldn't happen, panic

docgen/parser/parser.go
```diff
@@ -90,9 +92,51 @@ func parseMessages(pkgName string, comments map[string]*Comment, messages []*goo
...
+
+	// once we've defined all Messages, populate the NestedMessages
+	// member of each message
+	for _, m := range res {
+		m.NestedMessages = populateNestedMessages(m, res)
+	}
 	return res
 }
 
+func populateNestedMessages(message *Message, messages map[string]*Message) []*Message {
```
mvdan: this needs a godoc - is the incoming map generally empty? what is it used for? why are we returning a slice instead of just telling the caller to use the map?

mgfcmartinez:
@mvdan thanks, improved the docs a bit- let me know if they're still unclear.

re: anonymous fields- i've actually run into that by accident. it causes a panic in `loader` (so, before it even hits docgen). can i handle that in a separate PR?

mvdan:
Of course - we can deal with anonymous fields later :)

Issue #240, opened 2019-08-13
test TestScripts/config_protoc is failing on macOS


favadi:
Revision: 636f76a60bcce64f4d27c0c9ee3c0c4a0a40f805

```
=== CONT  TestScripts/config_protoc
    --- FAIL: TestScripts/config_protoc (31.75s)
        testscript.go:308: 
            WORK=$WORK
            PATH=/Users/favadi/code/gunk/gunk/.cache:/Users/favadi/bin:/Users/favadi/go-dev-tools/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/MacGPG2/bin:/Users/favadi/go/bin:/usr/local/Cellar/go/1.12.7/libexec/bin
            HOME=/Users/favadi/code/gunk/gunk/.cache/home
            TMPDIR=$WORK/tmp
            devnull=/dev/null
            :=:
            exe=
            GOPROXY=http://127.0.0.1:58702/mod
            GONOSUMDB=*
            
            # TODO: use [!net] once it does something useful.
            # See https://github.com/rogpeppe/go-internal/issues/75. (0.000s)
            > [short] skip 'requires network access' 
            # Use a separate $HOME, to not reuse a cached protoc. (0.000s)
            > env HOME=$WORK/protoc_home
            # Specify a valid protoc version.
            #
            # Note that the parent gunkconfig uses an invalid
            # version but since it's overriden by any existing child config
            # (i.e. here and in the succeeding tests) nothing blows up (31.708s)
            > gunk generate ./version
            > exists protoc_home/.cache/gunk/protoc-v3.8.0
            FAIL: testdata/scripts/config_protoc.txt:14: $WORK/protoc_home/.cache/gunk/protoc-v3.8.0 does not exist
```

Reason: $WORK/protoc_home/.cache/gunk/protoc-v3.8.0 does not exist should be `$WORK /protoc_home/Library/Caches/gunk/protoc-v3.8.0` for macOS. 

Also in config_protoc.txt, line 19: 

> stderr 'error: downloading protoc: could not find platform linux-x86_64 release asset for "invalid"'

will always fail on !linux machines.

mvdan:
@favadi please remember to add your "fixes" line to the commit, otherwise the magic doesn't work.

PR #242, opened 2019-08-19
fix test config_protoc failures on non Linux OSes

favadi:
- passing user cache dir variable to testscript
- changes pathversion test to work correctly across OSes

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -8,11 +8,13 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"testing"
 
-	"github.com/gunk/gunk/generate"
 	"github.com/rogpeppe/go-internal/goproxytest"
 	"github.com/rogpeppe/go-internal/testscript"
+
+	"github.com/gunk/gunk/generate"
 )
 
 var write = flag.Bool("w", false, "overwrite testdata output files")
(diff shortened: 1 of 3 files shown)


favadi:
Fixes #240.

main_test.go
```diff
@@ -138,13 +140,33 @@ func TestScripts(t *testing.T) {
...
+	realUserCacheDir, err := os.UserCacheDir()
+	if err != nil {
+		return "", err
+	}
+	realHomeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimPrefix(realUserCacheDir, realHomeDir), nil
```
mvdan: is there a reason why you're using a relative path? if so, please document; if not, simplify.

main_test.go
```diff
@@ -138,13 +140,33 @@ func TestScripts(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
 	testscript.Run(t, testscript.Params{
 		Dir: filepath.Join("testdata", "scripts"),
 		Setup: func(e *testscript.Env) error {
 			e.Vars = append(e.Vars, "GOPROXY="+proxyURL)
 			e.Vars = append(e.Vars, "GONOSUMDB=*")
 			e.Vars = append(e.Vars, "HOME="+home)
+			e.Vars = append(e.Vars, "CACHEPATH="+cachePath)
```
mvdan: you already know HOME here, so you could make CACHEPATH an absolute dir. see https://godoc.org/github.com/rogpeppe/go-internal/testscript#Env. Then you don't need ` env CACHEDIR=$HOME/$CACHEPATH`.

favadi: we actually don't use above HOME variable in test but created a new one with

> env HOME=$WORK/protoc_home

and as we don't know value $WORK until the test run, I think we couldn't construct CACHEPATH as an absolute directory.

main_test.go
favadi: comment added.

main_test.go
mvdan: Hmm, you're right. I don't know who added the `env HOME=...`, but that's weird, because we already set up HOME here. You can clean that up here, or in another issue+PR; up to you.

We do know WORK here, though. See the testscript.Env link I pasted above.

mgfcmartinez: it's to prevent conflicts with the other test that downloads protoc (which also sets its own HOME): https://github.com/gunk/gunk/blob/d518a3cca47de41926d00552661c825816c2db41/testdata/scripts/download.txt#L5

PR #246, opened 2019-08-20
docgen: generate sample response

favadi:


docgen/parser/parser.go
```diff
@@ -220,16 +220,39 @@ func parseMethods(pkgName string, messages map[string]*Message, methods []*googl
...
 	return res, nil
 }
 
+func generateMethodExample(messages map[string]*Message, name string) ([]string, error) {
+	_, ok := messages[name]
+	if !ok {
+		return nil, nil
+	}
+	p := genJSONExample(messages, name)
+	b, err := p.MarshalJSON()
```
mgfcmartinez: `MarshalJSON` already does [indentation](https://github.com/gunk/gunk/blob/636f76a60bcce64f4d27c0c9ee3c0c4a0a40f805/docgen/parser/parser.go#L371), but it's formatted specifically to look good in a [curl example](https://github.com/gunk/gunk/blob/995c81378bdeb318699e4cf08a191b2367a6519b/testdata/scripts/extract.txt#L570). i haven't looked into it too much but do you think maybe we could add properties (e.g. prefix, indent) to properties so that we don't do the indentation twice?

testdata/scripts/extract.txt
```diff
@@ -825,7 +882,6 @@ curl -X PUT \
 
 ### Responses
 
-<!-- TODO: add example -->
```
mgfcmartinez: this can be handled in a separate PR but we should probably figure out how better to communicate that a response is blank

docgen/parser/parser.go
favadi: > i haven't looked into it too much but do you think maybe we could add properties

we could, but I tried to remove the custom indent here and the generated markdown still looks good, so I think we should just remove it here. 

Removed and updated the test.

testdata/scripts/extract.txt
favadi: maybe we should print "response body is empty" or something, I think we should do it later.

mvdan:
Looks like you're hitting https://github.com/gunk/gunk/issues/248. Let's get that fixed ASAP; will have a quick look.

docgen/parser/parser.go
mgfcmartinez: removing the custom indentation affected the formatting though (the closing brackets aren't aligned). [example](https://github.com/gunk/gunk/blob/7c2723774205c59d4c49461de99b2df34c80327d/testdata/scripts/extract.txt#L569)

i think your initial approach is better from the end user's perspective. we indent twice, but at least the formatting is as expected

docgen/parser/parser.go
```diff
@@ -220,16 +220,35 @@ func parseMethods(pkgName string, messages map[string]*Message, methods []*googl
...
+	if err != nil {
+		return nil, err
+	}
+	return strings.Split(string(b), "\n"), nil
```
mgfcmartinez: forgot to ask earlier- might be missing something, but why the call to `strings.Split`? if example is just a string we can do

````
```json
{{$m.Example}}
```
````

File: docgen/parser/parser.go (hunk 1 above)
Comment (favadi) [reply docgen/parser/parser.go]: I moved the indented codes out of MarshalJSON to avoid indenting twice.

File: docgen/parser/parser.go (hunk 3 above)
Comment (favadi) [reply docgen/parser/parser.go]: no idea why, removing.

File: testdata/scripts/extract_nested.txt [hunk 4]
```diff
@@ -1,139 +1,3 @@
-# the two main points to note in this test are that the Transaction struct's Amount
-# field is a nested message that itself contains a nested message (Currency),
-# and it also contains two instances of the Account struct.
-#
-# the generated markdown correctly displays the doubly nested message
-# and only defines the Account object once.
-gunk generate transactions.gunk
```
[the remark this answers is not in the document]
favadi: I was using txtar-c, txtar-x, to modifying the test, probably they messed up the order, reverted.

docgen/parser/parser.go
```diff
@@ -220,16 +220,39 @@ func parseMethods(pkgName string, messages map[string]*Message, methods []*googl
...
 			Response:  messages[m.GetOutputType()],
+			Example:   example,
 			Operation: extOp.(*options.Operation),
-			Request:   req,
 		}
 	}
 	return res, nil
 }
 
+func generateMethodExample(messages map[string]*Message, name string) (string, error) {
```
mgfcmartinez: nit: `generateResponseExample` with `Method` field named `ResponseExample` would be clearer (at this point we might want to make `Method`'s `Response` field a type of its own)

favadi: did that in latest commit.

favadi:
> Looks like you're hitting #248. Let's get that fixed ASAP; will have a quick look.

Test now passed.

donskifarrell:
@favadi Whats the hold up on getting this merged?

favadi:
> Whats the hold up on getting this merged?

Just waiting for review.

Issue #248, opened 2019-08-20
remove the code to figure out the "latest" version, as it frequently runs into API rate limits


ghazninattarshah:
CI build fails with ```gunk generate ./...
error: could not find platform release asset for ""``` 

Refer: https://travis-ci.org/openbank/openbank/builds/574247172

mvdan:
This is getting worse, probably because we have more tests now than before. I'll have a look.

favadi:
I'll create a PR to add rate limiting to mitigate this.

mvdan:
To quote the interesting bit from the PR:

> The real fix would be to pin gunk to a specific version of protoc by default, letting the user override that in their gunkconfig. It's only the API that has a rate limit, not the actual download link for the archive. We'd lose the ability to figure out what the "latest" version is, but I'd say that's not a big deal. I'd rather have the tool be deterministic and simple.

In other words, partially revert the work that Ken did a few weeks ago to make it figure out what the "latest" version is. I'm sure we can figure out the download URL for an archive, given just:

* the version name
* GOOS
* GOARCH

mvdan:
This is still pretty high priority, as it blocks other gunk work, if anyone wants to tackle it. Just add it as a gunk subtask to this sprint.

favadi:
> This is still pretty high priority, as it blocks other gunk work, if anyone wants to tackle it. Just add it as a gunk subtask to this sprint.

I'll give it a try.

PR #249, opened 2019-08-20
generate: give a better error if download hits a 403

mvdan:
Status codes in the 4xx range don't give an error, so we have to check
for them ourselves. Do that.

For #248.

### DIFF
--- a/generate/download.go
+++ b/generate/download.go
@@ -175,6 +175,12 @@ func githubAssets(repo, version string) (string, []githubAsset, error) {
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode >= 400 {
+		// This can easily happen if we make tons of requests to GitHub
+		// from one machine, as the IP hits their default rate limit.
+		return "", nil, fmt.Errorf("GET %s: %s", urlstr, http.StatusText(res.StatusCode))
+	}
+
 	var release struct {
 		Name   string        `json:"name"`
 		Assets []githubAsset `json:"assets"`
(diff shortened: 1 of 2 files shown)


generate/download.go
```diff
@@ -175,6 +175,12 @@ func githubAssets(repo, version string) (string, []githubAsset, error) {
 	}
 	defer res.Body.Close()
 
+	if res.StatusCode >= 400 {
+		// This can easily happen if we make tons of requests to GitHub
+		// from one machine, as the IP hits their default rate limit.
+		return "", nil, fmt.Errorf("GET %s: %s", urlstr, http.StatusText(res.StatusCode))
```
mgfcmartinez: perhaps:

```
       // see https://developer.github.com/v3/#rate-limiting
	var apiError struct {
		Message string `json:"message"`
	}
	if err := json.NewDecoder(res.Body).Decode(&apiError); err != nil {
		return "", nil, fmt.Errorf("decoding error response: %v", err)
	}
 	return "", nil, fmt.Errorf("GET %s: %s: %s", urlstr, http.StatusText(res.StatusCode), apiError.Message) 
```
or similar?

`error: downloading protoc: GET https://api.github.com/repos/ProtocolBuffers/protobuf/releases/latest: Forbidden` is still a little opaque

mvdan: *shrug* I thought about it, but at the same time I want to get rid of this code soon, so I thought it wasn't worth the effort. At that point we might as well use a github API client.

PR #250, opened 2019-08-21
add rate limiter to avoid hitting Github download limit

### LINKED ISSUE #248: remove the code to figure out the "latest" version, as it frequently runs into API rate limits

favadi:
Fixes #248.


favadi:
> The proper fix would be to not hit the GitHub API at al

How about we do a setup step in .travis.yaml that pre-download all necessary Github assets and serve it locally? We will need to make the CheckOrDownloadProtoc to an interface that either fetch from Github, or with local files in test.

mvdan:
That could fix the CI, but it still wouldn't fix the real problem. And we also wouldn't test if our real code querying GitHub would break. So the result would be more brittle, and users would still be able to hit a rate limit.

The real fix would be to pin gunk to a specific version of protoc by default, letting the user override that in their gunkconfig. It's only the API that has a rate limit, not the actual download link for the archive. We'd lose the ability to figure out what the "latest" version is, but I'd say that's not a big deal. I'd rather have the tool be deterministic and simple.

favadi:
> It's only the API that has a rate limit, not the actual download link for the archive

I didn't realize that, closing the PR as we need to take an completely different approach.

PR #251, opened 2019-08-21
docgen: remove support for downloading latest version of protoc

### LINKED ISSUE #248: remove the code to figure out the "latest" version, as it frequently runs into API rate limits

favadi:
to avoid hitting Github API rate limit

Fix #248

### DIFF
--- a/generate/download.go
+++ b/generate/download.go
@@ -3,20 +3,21 @@ package generate
 import (
 	"archive/zip"
 	"bytes"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"path/filepath"
-	"regexp"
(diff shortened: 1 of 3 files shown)


favadi:
> Perhaps include context in the commit message, as otherwise a git log reader doesn't know why this is being removed.

I did state the reason is "to avoid hitting Github API rate limit" in git commit message, do you think we will need more details?

mvdan:
Yes, I think a bit more than a short sentence would be good :)

favadi:
> Yes, I think a bit more than a short sentence would be good :)

commit message updated.

PR #254, opened 2019-08-28
fix infinite loop when generate unresolvable import

favadi:
when executing "generate" if an import package is unresolvable, gunk
creates infinite number of gunpkg-*.go files until terminated.
This happens because a recursive calls in Loader.Loop method that
occurs when check.Files returns an error.

Solution: terminate loader.Load if an error is reported.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -349,6 +349,11 @@ func (l *Loader) parseGunkPackage(pkg *GunkPackage) {
 		pkg.ProtoName = pkg.Name
 	}
 
+	// the reported error will be handle at generate.Run function.
+	if len(pkg.Errors) > 0 {
+		return
+	}
+
 	if !l.Types {
 		return
 	}
--- a/testdata/scripts/loader_files.txt
+++ b/testdata/scripts/loader_files.txt
@@ -13,7 +13,7 @@ cmp p2/p2.gunk p2/p2.gunk.golden
 
 # loading files from different packages with types should error
 ! gunk generate p1/p1.gunk p2/p2.gunk
-stderr 'package p2; expected p1'
+stderr '-: gunk package name mismatch: "p1" "p2"'
 
 -- go.mod --
 module testdata.tld/util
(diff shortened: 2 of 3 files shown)


testdata/scripts/loader_files.txt
```diff
@@ -13,7 +13,7 @@ cmp p2/p2.gunk p2/p2.gunk.golden
 
 # loading files from different packages with types should error
 ! gunk generate p1/p1.gunk p2/p2.gunk
-stderr 'package p2; expected p1'
+stderr '-: gunk package name mismatch: "p1" "p2"'
```
mvdan: hmm, why is this changed?

loader/loader.go
```diff
@@ -349,6 +349,10 @@ func (l *Loader) parseGunkPackage(pkg *GunkPackage) {
 		pkg.ProtoName = pkg.Name
 	}
 
+	if len(pkg.Errors) > 0 {
+		return
```
mvdan: shouldn't we report the errors? if they are caught elsewhere, just mention that in a comment here.

favadi: comment added.

testdata/scripts/loader_files.txt
favadi: because with the change in loader.go, we now catch/return this error earlier.

testdata/scripts/generate_unexisting_import.txt
```diff
@@ -0,0 +1,14 @@
+# make sure go modules is turned off, and GOPATH is set to a nonexistent directory
```
mvdan: Just noticed "unexisting" in this file name - the correct term would be "nonexistent"

mvdan: Or even better, "missing" :)

favadi: > Just noticed "unexisting" in this file name - the correct term would be "nonexistent"
 
changed all to nonexistent.

PR #255, opened 2019-08-29
update docgen  for dependency importing

kofoworola:
Edited the docgen plugin to properly add the documentation for imported Messages (structs). Also fixed a bug where docgen didn't generate for files that have imports

### DIFF
--- a/docgen/docgen.go
+++ b/docgen/docgen.go
@@ -9,10 +9,10 @@ import (
 	"strings"
 
 	"github.com/golang/protobuf/proto"
-	google_protobuf "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	plugin_go "github.com/golang/protobuf/protoc-gen-go/plugin"
 
 	"github.com/gunk/gunk/docgen/generate"
+	"github.com/gunk/gunk/docgen/parser"
 	"github.com/gunk/gunk/log"
 	"github.com/gunk/gunk/plugin"
 )
@@ -52,19 +52,23 @@ func (p *docPlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.Co
 		}
 	}
 
-	var source *google_protobuf.FileDescriptorProto
+	// find the source by looping through the protofiles and finding the one that matches the FileToGenerate
+	var source *parser.FileDescWrapper
 	for _, f := range req.GetProtoFile() {
 		if strings.Contains(f.GetName(), "all.proto") {
-			source = f
-			break
+			for _, fileToGenerate := range req.FileToGenerate{
+				if fileToGenerate == f.GetName(){
+					source = &parser.FileDescWrapper{FileDescriptorProto: f}
+					break
+				}
+			}
 		}
 	}
-
 	if source == nil {
 		return nil, fmt.Errorf("no file to generate")
 	}
-
 	base := filepath.Join(filepath.Dir(source.GetName()))
+	source.DependencyMap = parser.GenerateDependencyMap(source.FileDescriptorProto,req.GetProtoFile())
 
 	var buf bytes.Buffer
 	pb, err := generate.Run(&buf, source, lang)
--- a/docgen/parser/model.go
+++ b/docgen/parser/model.go
@@ -1,8 +1,15 @@
 package parser
 
 import (
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
 )
+
+// FileDescWrapper is a wrapper for FileDescriptorProto for holding a map of dependencies.
+type FileDescWrapper struct {
+	*descriptor.FileDescriptorProto
+	DependencyMap map[string]*descriptor.FileDescriptorProto
+}
 
 // File is a proto parsed file.
 type File struct {
--- a/docgen/parser/parser.go
+++ b/docgen/parser/parser.go
@@ -27,11 +27,14 @@ const (
 )
 
 // ParseFile parses a proto file.
-func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
+func ParseFile(file *FileDescWrapper) (*File, error) {
 	comments := parseComments(file.GetSourceCodeInfo())
 
 	pkgName := file.GetPackage()
-	messages := parseMessages(pkgName, comments, file.GetMessageType())
+	messages, err := nestedDescriptorMessages(*file.FileDescriptorProto, file.DependencyMap)
+	if err != nil {
+		return nil, err
+	}
 	services, err := parseServices(pkgName, messages, file.GetService())
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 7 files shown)


mvdan:
As mentioned on CI, this isn't ready; please ping once it is.

donskifarrell:
@kofoworola Lets try to get this resolved + merged before sprint closes :) 

@mgfcmartinez added you as a reviewer due to your closeness to the code here.

docgen/docgen.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/golang/protobuf/proto"
 	google_protobuf "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	plugin_go "github.com/golang/protobuf/protoc-gen-go/plugin"
+	"github.com/gunk/gunk/docgen/parser"
```
mvdan: gunk imports go below.

docgen/docgen.go
```diff
@@ -52,18 +53,33 @@ func (p *docPlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.Co
...
 	}
 
-	var source *google_protobuf.FileDescriptorProto
+	var source *parser.FileDescWrapper
 	for _, f := range req.GetProtoFile() {
 		if strings.Contains(f.GetName(), "all.proto") {
-			source = f
-			break
+			for _, fileToGenerate := range req.FileToGenerate {
+				if fileToGenerate == f.GetName() {
```
mvdan: why is the extra loop necessary? you don't even use fileToGenerate for anything, as far as I can see

docgen/docgen.go
```diff
@@ -52,18 +53,33 @@ func (p *docPlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.Co
...
 		}
 	}
 
 	if source == nil {
 		return nil, fmt.Errorf("no file to generate")
 	}
 
+	source.DependencyMap = make(map[string]*google_protobuf.FileDescriptorProto)
+
+	// generate dependency array for the source
```
mvdan: sorry, what do you mean by "array"? I just see a map :)

docgen/parser/model.go
```diff
@@ -1,9 +1,16 @@
 package parser
 
 import (
+	"github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
 )
 
+// FileDescWrapper is a wrapper for FileDescriptorProto for holding a map of dependencies
```
mvdan: missing period

docgen/parser/parser.go
```diff
@@ -54,6 +58,60 @@ func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
...
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func fieldPackage(typeName string) string {
+	//lastDot := strings.LastIndex(typeName,".")
```
mvdan: remove this

docgen/parser/parser.go
```diff
@@ -54,6 +58,60 @@ func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
 	return f, nil
 }
 
+func nestedDescriptorMessages(file google_protobuf.FileDescriptorProto, genFiles map[string]*google_protobuf.FileDescriptorProto) (map[string]*Message, error) {
```
mvdan: document, please

docgen/parser/parser.go
```diff
@@ -27,11 +27,15 @@ const (
 )
 
 // ParseFile parses a proto file.
-func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
+func ParseFile(file *FileDescWrapper) (*File, error) {
 	comments := parseComments(file.GetSourceCodeInfo())
 
 	pkgName := file.GetPackage()
-	messages := parseMessages(pkgName, comments, file.GetMessageType())
+	//nested,err := nestedDescriptorMessages(file, dependencies)
```
mvdan: remove this too. please try to not keep commented out code in PRs, unless there's a good reason for it

docgen/parser/parser.go
```diff
@@ -54,6 +58,60 @@ func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
...
+				if fieldPackage(field.GetTypeName()) == *dep.Package {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func fieldPackage(typeName string) string {
```
mvdan: what does this function do? the name and code aren't terribly obvious.

docgen/parser/parser.go
```diff
@@ -54,6 +58,60 @@ func ParseFile(file *google_protobuf.FileDescriptorProto) (*File, error) {
...
+	dependencies := file.GetDependency()
+	pkgName := file.GetPackage()
+	comments := parseComments(file.GetSourceCodeInfo())
+	for _, dep := range dependencies {
+		if _, ok := genFiles[dep]; !ok {
+			// not needed
+			continue
+		}
+		// check if dependency is needed for messages
+		if isDepNeeded(file, genFiles[dep]) {
```
mvdan: invert and early continue

mvdan:
Diep can also help review this :)

docgen/parser/parser.go
kofoworola: Oh, i haven't cleaned it up yet and removed the comments. Would do so now. Just wanted to make sure it worked fo @favadi

mvdan: Well, you did clearly ask for a review+merge ;)

mvdan:
Also, this needs a test. @favadi has written multiple test scripts in gunk, so he can help you. You can just take one of the recent ones and adapt it to test this fix or feature.

docgen/parser/parser.go
mvdan: Sorry, wrong PR!

docgen/docgen.go
kofoworola: This checks if the file in the outer loop `f` is part of the descriptors filesToGenerate. Before this, it simply chose the first in the ProtoFile, which is usually a dependency in this case @mvdan

mvdan: Ok, gotcha - please document this.

mgfcmartinez: i would suggest 

```go
var source *parser.FileDescWrapper
for _, ftg := range req.FileToGenerate {
	if !strings.Contains(f.GetName(), "all.proto") {
             continue
        }
        source = f
        break
}
```
no need to iterate over ProtoFile at this point, since you're just looking for the file we actually want to generate

https://godoc.org/github.com/golang/protobuf/protoc-gen-go/plugin#CodeGeneratorRequest

docgen/parser/parser.go
kofoworola: Sorry... i'm confused 😅

generate/generate.go
```diff
@@ -440,7 +440,7 @@ func (g *Generator) translatePkg(pkgPath string) error {
 
 // fileOptions will return the proto file options that have been set in the
 // gunk package. These include "JavaPackage", "Deprecated", "PhpNamespace", etc.
-func (g *Generator) fileOptions(pkg *loader.GunkPackage) (*desc.FileOptions, error) {
+func FileOptions(pkg *loader.GunkPackage) (*desc.FileOptions, error) {
```
mgfcmartinez: was this left in accidentally? doesn't seem to be needed outside of `generate`

docgen/docgen.go
kofoworola: Oh ok thanks for this @mgfcmartinez

docgen/docgen.go
```diff
@@ -52,18 +53,33 @@ func (p *docPlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.Co
...
+	for _, d := range source.GetDependency() {
```
mgfcmartinez: i'm on the fence about this- on one hand this is probably safer, but on the other, i think we can also expect the `ProtoFile` contents to be accurate (i.e. it contains decriptors for the contents of `FileToGenerate` + their imports), since we handle that in gunk proper. if you want to leave this in, i would also check to make sure that every element in from `GetDependency` was present

docgen/parser/parser.go
mgfcmartinez:
```go
if !isDepNeeded(file, genFiles[dep]) {
     continue
}

dependencyMessages, err := nestedDescriptorMessages(*genFiles[dep], genFiles)
if err != nil {
    return nil, err
}

. . .
```

to decrease the indentation level of the meat of the lgoic

kofoworola: oh, ok thanks for this

generate/generate.go
kofoworola: Yes, it was, thanks for catching it. I needed it to write tests

docgen/docgen.go
kofoworola: changed the comment

docgen/parser/parser.go
kofoworola: Added a short description, please let me know if it needs more context

docgen/docgen.go
kofoworola: @mgfcmartinez the ftg, in this case, will be a string but the source has to be/have a FileDescriptorProto though.

mgfcmartinez: @kofoworola right- my bad. you can iterate over req.ProtoFile instead, since it should also contain the file to generate. you could also take the last element in the slice, since we [sort](https://github.com/gunk/gunk/blob/4587cc79bbf6903ef45ef837099dbf3714cde8ad/generate/generate.go#L320) it beforehand

docgen/docgen.go
kofoworola: The reason i opted to loop through the source's dependency list is the requests ProtoFile contained a lot of unneeded files (e.g annotation parsers) in some cases

docgen/docgen.go
kofoworola: I think taking the last element is better, skipped my mind topologicalSort already sorts it

kofoworola: but then again there needs to be check for if the FilesToGenerate is empty, so i would leave it

kofoworola: Also @mgfcmartinez after running it a few times, there were some scenarios where the FileToGenerate was not the last protofile. Perhaps i could create an issue for that later on

docgen/docgen.go
mvdan: @mgfcmartinez if you still think we can do this better, please send a follow-up PR. I don't want to hold this for much longer.

PR #256, opened 2019-08-29
fix gunk convert panic when option contains []byte value

### LINKED ISSUE #245: gunk convert panic: string is not a valid value for uint8 for openapiv2_schema option
gunk convert panics when running with following valid proto file. 

Reproduce testscript:

```txt
gunk convert util.proto

-- util.proto --
syntax = "proto3";

package util;

message Util {
    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
	  example: {
	    value: '{ "status": "ok" }'
	  }
    };

    string hello = 1;
}
```

favadi:
Fix #245

### DIFF
--- a/reflectutil/setter.go
+++ b/reflectutil/setter.go
@@ -125,19 +125,24 @@ func valueFor(typ reflect.Type, tag reflect.StructTag, value interface{}) reflec
 		}
 		return mp
 	case reflect.Slice:
+		etyp := typ.Elem()
 		list := reflect.MakeSlice(typ, 0, 0)
 		switch value := value.(type) {
 		case *ast.CompositeLit:
 			for _, elt := range value.Elts {
-				list = reflect.Append(list, valueFor(typ.Elem(), tag, elt))
+				list = reflect.Append(list, valueFor(etyp, tag, elt))
 			}
 		case *protop.Literal:
+			// convert string to slice of bytes, uint8 and byte are the same kind
+			if etyp.Kind() == reflect.Uint8 && value.IsString {
+				return reflect.ValueOf([]byte(value.SourceRepresentation()))
+			}
 			if value.Array == nil {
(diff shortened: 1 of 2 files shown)


reflectutil/setter.go
```diff
@@ -155,6 +162,8 @@ func valueFor(typ reflect.Type, tag reflect.StructTag, value interface{}) reflec
 		valueStr = x.Sel.Name
 	case *protop.Literal:
 		valueStr = x.SourceRepresentation()
+	case uint8:
+		return reflect.ValueOf(x)
```
mvdan: this case stands out a bit from the rest. add a comment?

reflectutil/setter.go
```diff
@@ -132,6 +132,13 @@ func valueFor(typ reflect.Type, tag reflect.StructTag, value interface{}) reflec
 				list = reflect.Append(list, valueFor(typ.Elem(), tag, elt))
 			}
 		case *protop.Literal:
+			if typ.Elem().Kind() == reflect.TypeOf(byte(0)).Kind() && value.IsString {
```
mvdan:
```
if etyp := typ.Elem(); etyp.Kind() == reflect.Uint8 && value.IsString {
```

(and reuse `etyp` below)

mvdan: Or even better, declare `etyp` at the top of the `case reflect.Slice:`

reflectutil/setter.go
favadi: yeah, that doesn't look right, will trying to make it more obvious.

reflectutil/setter.go
favadi: > Or even better, declare etyp at the top of the case reflect.Slice:

I did that in latest commit.

reflectutil/setter.go
favadi: I removed this special case and return the whole slice of bytes instead.

PR #261, opened 2019-09-13
fixed crash when encountering anonymous struct attributes

kofoworola:
Made gunk reject anonymoius/unamed attributes and added a human friendly error message when encountered. Also ran gofmt

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -405,6 +405,14 @@ func (l *Loader) validatePackage(pkg *GunkPackage) {
 				return true
 			}
 
+			// Look through all fields for anonymous/unnamed types.
+			for _, field := range st.Fields.List {
+				if len(field.Names) < 1 {
+					pkg.addError(ParseError, st.Pos(), l.Fset, "anonymous struct fields are not supported")
+					return false
+				}
+			}
+
 			// Check for struct tag 'pb' and ensure that if it does exist
 			// it is a valid integer, and it is unique in that struct.
 			// The other validation should happen in format and generate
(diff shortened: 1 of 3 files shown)


loader/loader.go
```diff
@@ -405,6 +405,14 @@ func (l *Loader) validatePackage(pkg *GunkPackage) {
 				return true
 			}
 
+			//look through all fields for anonymous/unnamed types
```
mvdan: space this comment, capitalise it, and add a period.

loader/loader.go
```diff
@@ -405,6 +405,14 @@ func (l *Loader) validatePackage(pkg *GunkPackage) {
...
+			for _, field := range st.Fields.List {
+				if field.Names == nil || len(field.Names) < 1 {
```
mvdan: remove the nil check.

loader/loader.go
```diff
@@ -405,6 +405,14 @@ func (l *Loader) validatePackage(pkg *GunkPackage) {
...
+					pkg.addError(ParseError, st.Pos(), l.Fset, "anonymous struct attributes are not supported")
```
mvdan: s/attributes/fields/

mvdan:
Right, I completely forgot to spot that this is missing a test. You've written test scripts before, right? Please write a simple one for this crash. There might already be a testscript that covers a bunch of previous crashes, in which case just add to it.

kofoworola:
@mvdan @mgfcmartinez added the test script. Kindly re-review

PR #264, opened 2019-09-26
[DRAFT] generate: replace field names from UpperCamelCase to snake_case

### LINKED ISSUE #263: generate: `name` and `json` file descriptor are not in the standard snake_case naming convention
When I was trying to generate the following code from gunk, the resulting proto was not as expected. The example below illustrates the part.

gunk code:
```
type CreateTransferResponse struct {
    TransferID string `pb:"1" json:"transfer_id"`
}
```

generated artifact:
```
type CreateTransferResponse struct {
    TransferID string `protobuf:"bytes,1,opt,name=TransferID,json=transfer_id,proto3" json:"TransferID,omitempty"`
    XXX_NoUnkeyedLiteral struct{} `json:"-"`
    XXX_unrecognized     []byte   `json:"-"`
    XXX_sizecache        int32    `json:"-"`
}
```

I was expecting to see:
```
    TransferID string `protobuf:"bytes,1,opt,name=transfer_id,json=transfer_id,proto3" json:"transfer_id,omitempty"`
```

Which should be as per protobuf naming convention of using snake_case.

jwangsadinata:
We used UpperCamelCase when writing variable/field names in Go, but the
convention for writing a protobuf file is to use a snake_case, such
as the following:

```
message SomeMessage {
    string user_id = 1
}
```

Currently, the `generate` code just copies the field name, which results
in this proto file before outputting the generated artifact:

```
message SomeMessage {
    string UserID = 1
}
```

which causes issues with the json tag on the file descriptor

This commit fixes the above by replacing the field name to snake_case,
per protobuf specification.

Fixes #263.


mvdan:
This must have been a regression, right? I imagine we haven't done this wrong from the beginning.

Also, this needs a test.

generate/generate.go
```diff
@@ -653,7 +654,7 @@ func (g *Generator) convertMessage(tspec *ast.TypeSpec) (*desc.DescriptorProto,
 		if len(field.Names) != 1 {
 			return nil, fmt.Errorf("need all fields to have one name")
 		}
-		fieldName := field.Names[0].Name
+		fieldName := snaker.CamelToSnake(field.Names[0].Name)
```
favadi: hmm, not sure if I understand this. For example 

```
type Test1 struct {
  Field1 string `pb:"1" json:"field2_yes_not_a_typo"`
}
```

I would expect the generated .go file has

```
type Test1 struct {
  Field 1 string `... json:"field2_yes_not_a_typo,omitempty"`
}
```

That means we shouldn't need to do any case transformation.

I haven't really look at the surrounding codes, just this line, but this does need explanation.

jwangsadinata:
Hi @mvdan and @favadi , thanks so much for jumping in. Definitely need some feedback on my approach too, and will be writing some tests before taking the `DRAFT` off.

Anyway, I went through the `generate` code, and it seems the steps are converting the `gunk` definition into `proto`, and then running it against the different generators.

I looked at this: https://developers.google.com/protocol-buffers/docs/proto3, and the part that made me did this was:

> Message field names are mapped to lowerCamelCase and become JSON object keys. If the json_name field option is specified, the specified value will be used as the key instead. Parsers accept both the lowerCamelCase name (or the one specified by the json_name option) and the original proto field name. null is an accepted value for all field types and treated as the default value of the corresponding field type.

Since all the examples and naming conventions of the proto file are in snake_case, what I did was basically converting all gunk variable names from UpperCamelCase, to it's snake case counterpart.

Currently, if we do a generate from the following:
```
type SomeMessage struct {
  UserID string `pb:"1" json:"user_id"`
}
```

we'll get something like the following as the proto (which is unexposed)
```
message SomeMessage {
    string UserID = 1
}
```

What I wanted, is to just change the field name as well, so that we have:
```
message SomeMessage {
    string user_id = 1
}
```

I might be approaching this thing in a wrong way though. Might just look into how I could override the json output. Let me know if you have some opinions @favadi @mvdan.

jwangsadinata:
btw @mvdan, just for context, the goal for this PR is just to have a regular json output like this:

    {"msg": "bleh", "code": 2}

instead of:

    {"Msg": "bleh", "Code": 2}

which is currently the case.

jwangsadinata:
I am closing this PR since I could use the `runtime.WithMarshalerOption` on grpc-gateway to override the json tag that is generated. Changing the whole PR will be quite a redesign on how gunk generate works, and it is not quite necessary.

However, it's worth pointing out that there are still differences between gunk generated code, and protobuf generated code, as detailed in https://github.com/gunk/gunk/issues/33.

PR #266, opened 2019-10-05
format: allow receiving files from different packages

### LINKED ISSUE #265: format: allow receiving files from different packages
Right now it complains if the package names don't match. That's silly; formatting is on a per-file basis.

We work around this by alling `gunk format file.go` for each file separately, which is quite a bit slower than calling `gunk format` once with all the files as arguments.

Should be a simple fix, I think. A test would also be pretty easy.

Oiyoo:
creates a new error kind for proto name mismatch

adds a helper func to remove the error kind from []*loader.GunkPackage

fixes #265

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -339,7 +339,7 @@ func (l *Loader) parseGunkPackage(pkg *GunkPackage) {
 		}
 		if pkg.ProtoName == "" {
 			pkg.ProtoName = name
-		} else if name != "" {
+		} else if name != "" && l.Types {
 			pkg.addError(ValidateError, 0, nil, "proto package name mismatch: %q %q",
 				pkg.ProtoName, name)
 			continue
--- a/testdata/scripts/format.txt
+++ b/testdata/scripts/format.txt
@@ -12,6 +12,11 @@ cd broken
 stderr 'echo.gunk:1'
 cd ..
 
+# formatting files from different packages and checking the result
+gunk format echo.gunk ./diffpkg/notutil.gunk
+cmp echo.gunk echo.gunk.golden
+cmp ./diffpkg/notutil.gunk notutil.gunk.golden
+
 -- go.mod --
 module testdata.tld/util
 -- empty/empty.go --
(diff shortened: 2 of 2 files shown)


loader/loader.go
```diff
@@ -340,7 +341,7 @@ func (l *Loader) parseGunkPackage(pkg *GunkPackage) {
 		if pkg.ProtoName == "" {
 			pkg.ProtoName = name
 		} else if name != "" {
```
mvdan: I had a much simpler solution in mind, which perhaps I should have clarified on the ticket.

We can simply skip the package name mismatch check if `l.Types == false`. We don't need to typecheck, so we don't need to follow the restriction.

In technical terms a package should still contain only the files for that package. But in practice, we don't care all that much, as the only commands that load packages are format (syntax-only and per-file) and generate (with `Types == true`). If we need the current behavior in the future (correct packages with sanity checks, but no type-checking), we can always add more knobs to Loader.

mvdan: This also means that you can remove much of the added code in format.go.

Oiyoo: done, thanks!

donskifarrell:
@Oiyoo Any luck on that test case? I'm sure there are other devs that can help review who have experience with Gunk so we don't wait until Dan is back online

mvdan:
We discussed this privately on Slack, so my bad for not updating this thread.

I wanted to do #267 first, as I thought gunk's tests wouldn't pass on the current version of Go otherwise. But I think I'm wrong. @Oiyoo feel free to add the test now. If you encounter failures with Go 1.13.x, just download and use Go 1.12.x.

donskifarrell:
@Oiyoo Can we get this merged soon?

testdata/scripts/format.txt
```diff
@@ -5,6 +5,7 @@ stderr 'no Gunk packages to format'
 # formatting a normal package and checking the result
 gunk format .
 cmp echo.gunk echo.gunk.golden
+cmp notutil.gunk notutil.gunk.golden
```
mvdan: you just compare the files? I think you're missing the actual `gunk format` call with files.

also, separate the test from the "formatting a normal package" via an empty line and another section comment.

Oiyoo: danke, sir

PR #269, opened 2019-10-11
scopegen: remove commas when generating Go const fields

cadaroya:
When generated, current Go const values are separated by commas.
This causes the generated all.scopes.go file to not work properly due to syntax error.
This was easily fixed by removing the comma in the generate file.

### DIFF
--- a/testdata/scripts/scopegen.txt
+++ b/testdata/scripts/scopegen.txt
@@ -1,4 +1,5 @@
 gunk generate -v echo.gunk
+exec go build all.scopes.go
 cmp all.scopes.json all.scopes.json.golden
 cmp all.scopes.go all.scopes.go.golden
 -- go.mod --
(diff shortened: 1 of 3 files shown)


testdata/scripts/scopegen.txt
```diff
@@ -88,9 +88,9 @@ package test
 type AuthScope string
 
 const (
-	Scope_admin AuthScope = "Grants read and write access to administrative information",
-	Scope_read AuthScope = "Grants read access",
-	Scope_write AuthScope = "Grants write access",
+	Scope_admin AuthScope = "Grants read and write access to administrative information"
+	Scope_read AuthScope = "Grants read access"
+	Scope_write AuthScope = "Grants write access"
```
favadi: could you add a step in this script to make sure the generated go file can compile? Otherwise, LGTM.

cadaroya: @favadi i added an `exec go build all.scopes.go` cmd to check for the syntax. hope this solves it!

PR #272, opened 2019-10-26
scopegen: create standardized scope challenges

Kunde21:
Using a set of authorized scopes, ensure the request is Authorized by
testing against the method's defined scopes.

In order to ensure the method scopes are not changed during execution,
the method scopes are unexported.

Usage:  Embed the <package>.ServiceScope type in the service struct,
then use the `Any` or `All` challenges in an interceptor, with the set
of user authorized scopes.


scopegen/generate/go.go
```diff
@@ -20,11 +21,44 @@ const (
...
-var AuthScopes = map[string][]AuthScope{
+var authScopes = map[string][]AuthScope{
 {{- range $method := .Methods }}
 	"{{ $method.Name }}": []AuthScope{ {{- range $index, $scope := $method.Scopes }}{{ if $index }}, {{ end }}Scope_{{ $scope }}{{ end -}} },
 {{- end }}
 }
+
+// Any allows a loose challenge, for claims containing any of the method scopes.
+//
+// Use All method as a defualt for OAuth2 style scopes.  Any is usful with more complex scope definitions.
```
favadi: defualt --> default.
usful --> useful.

scopegen/generate/go.go
```diff
@@ -20,11 +21,44 @@ const (
...
+func (svcSc ServiceScope) Any(method string, claims []string) bool {
```
favadi: maybe just make Any/All a global function? I don't see why we need ServiceScope struct.

favadi:
Also, CI is failing.

scopegen/generate/go.go
Kunde21: The ServiceScope struct allows it to be passed as an interface.  All services would be able to define their scopes using `<svc>.ServiceScope` as an embedded field.  Interceptors can use that interface via type assertion, with minimal changes to the service implementation itself.

favadi: Could you add an code example somewhere in README.md for this?

Kunde21: I plan on creating a basic interceptor, using that to show its use.

PR #275, opened 2020-01-08
docgen: fix error when encountering map type

kofoworola:
docgen breaks when encountering a type of map[string/int]value. This commit fixes it by extracting the map fields of a message from the NestedType field of DescriptorProto struct.

### DIFF
--- a/docgen/parser/parser.go
+++ b/docgen/parser/parser.go
@@ -155,14 +155,69 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
 	return res
 }
 
+// extractMapFields generates an array of fields that correspond to map type fields for messages, which are stored in
+// the NestedType field of the DescriptorProto struct
+func extractMapFields(message *google_protobuf.DescriptorProto, path string, comments map[string]*Comment) []*Field {
+	fields := make([]*Field, len(message.NestedType))
+	for i, n := range message.GetNestedType() {
+		// a nested type of map will always have two fields for key and value
+		if !n.GetOptions().GetMapEntry(){
+			continue
+		}
+		if len(n.GetField()) != 2 {
+			panic("a map type message should always have two fields")
+		}
+
+		nestedTypeName := strings.TrimSuffix(n.GetName(), "Entry")
+		// find position and fieldDescriptorProto of the map type from the message fields
+		pos := -1
+		var fieldDescriptorProto *google_protobuf.FieldDescriptorProto
+		for i, f := range message.GetField() {
+			if f.GetName() == nestedTypeName {
+				pos = i
+				fieldDescriptorProto = f
+				break
+			}
+		}
+		if pos == -1 || fieldDescriptorProto == nil {
+			continue
+		}
+
+		// get key type. key can only be string or int32
+		var keyType string
+		switch n.GetField()[0].GetType() {
+		case google_protobuf.FieldDescriptorProto_TYPE_INT32:
+			keyType = "int"
+		case google_protobuf.FieldDescriptorProto_TYPE_STRING:
+			keyType = "string"
+		default:
+			panic("key should either be a string or an int32")
+		}
+
+		field := Field{
+			Name:    nestedTypeName,
+			Comment: nonNilComment(comments[fmt.Sprintf("%s.%d.%d", path, fieldFlag, pos)]),
+			Type: &Type{
+				Name:          fmt.Sprintf("map[%s]%s", keyType, getType(n.GetField()[1]).Name),
+				IsArray:       true,
+			},
+		}
+		fields[i] = &field
+	}
+	return fields
+}
+
 func parseMessages(pkgName string, comments map[string]*Comment, messages []*google_protobuf.DescriptorProto, merge map[string]*Message) map[string]*Message {
 	// convert each proto message to our Message representation
 	for i, m := range messages {
 		p := fmt.Sprintf("%d.%d", messageFlag, i)
+		mapFields := extractMapFields(m, p, comments)
+		fields := parseFields(p, comments, m.GetField())
+		fields = append(fields, mapFields...)
 		merge[getQualifiedName(pkgName, m.GetName())] = &Message{
 			Name:    m.GetName(),
 			Comment: nonNilComment(comments[p]),
-			Fields:  parseFields(p, comments, m.GetField()),
+			Fields:  fields,
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


greywolve:
I think @mvdan or @Kunde21 are better positioned to review this than me. My only comment is that I'd probably do `extractMapFields` inside `parseFields` that seems like a more straight forward way of understanding what's happening rather than having to do a special extra extract step, and then still do parse fields too. Otherwise, nice work :)

kofoworola:
just came across another bug when testing, not sure if it's related to this. So I'll update the PR with it once done.

kofoworola:
Updated the bug fix

kofoworola:
@greywolve the reason why extracMap is in another step is that it does not work with the `fields` parameters passed into parseField and it needs access to the parent message itself.

testdata/scripts/generate_docgen_map.txt
```diff
@@ -0,0 +1,82 @@
...
+package docgentest
+
+// Account holds all details about a bank account.
+type Sample struct {
+    // MapStringInt is a simple map of a <string,int>.
+	MapStringInt map[string]int `pb:"1" json:"map_string_int"`
+	// MapStringInt is a simple map of a <int,string>.
+	MapIntString map[int]string `pb:"2" json:"map_int_string"`
+	// MapStringPrice is a sample map of <string,price>
+	MapStringPrice map[string]Price `pb:"3" json:"map_string_price"`
```
mvdan: please use consistent formatting, ideally with tabs

docgen/parser/parser.go
```diff
@@ -219,14 +268,19 @@ func populateNestedMessages(message *Message, messages map[string]*Message) {
 }
 
 func parseFields(path string, comments map[string]*Comment, fields []*google_protobuf.FieldDescriptorProto) []*Field {
-	res := make([]*Field, len(fields))
+	var res []*Field
 	for i, f := range fields {
-		res[i] = &Field{
+		// check if this field is a map which should be handled by the extractMapField
+		if strings.Contains(f.GetTypeName(), f.GetName()+"Entry") {
```
mvdan: This seems a bit fragile. What if a type is really just named `FooEntry` because the user called it that?

It seems like you could use `MessageOptions.MapEntry`, which is a `*bool`.

docgen/parser/parser.go
```diff
@@ -155,14 +155,63 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
...
+		if n.GetField()[0].GetType() == google_protobuf.FieldDescriptorProto_TYPE_INT32 {
+			keyType = "int"
+		}
+		field := Field{
+			Name:    nestedTypeName,
+			Comment: nonNilComment(comments[fmt.Sprintf("%s.%d.%d", path, fieldFlag, pos)]),
+			Type: &Type{
+				Name:          fmt.Sprintf("map[%s]%s", keyType, getType(n.GetField()[1]).Name),
+				QualifiedName: "",
+				IsMessage:     false,
```
mvdan: zero values should be left out.

docgen/parser/parser.go
```diff
@@ -155,14 +155,63 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
...
 }
 
+// extractMapFields generates an array of fields that correspond to map type fields for messages, which are stored in
+// the NestedType field of the DescriptorProto struct
+func extractMapFields(message *google_protobuf.DescriptorProto, path string, comments map[string]*Comment) []*Field {
+	fields := make([]*Field, len(message.NestedType))
+	for i, n := range message.GetNestedType() {
+		// a nested type of map will always have two fields for key and value
+		if len(n.GetField()) != 2 {
+			continue
```
mvdan: this should be a panic, because of the comment above. you can also just make the comment be the panic message, and simplify the code.

if you do a `continue`, we might be silently ignoring data.

docgen/parser/parser.go
```diff
@@ -155,14 +155,63 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
...
+				pos = i
+				fieldDescriptorProto = f
+				break
+			}
+		}
+		if pos == -1 || fieldDescriptorProto == nil {
+			continue
+		}
+
+		// get key type. key can only be string or int32
```
mvdan: if this is the case, again, please do a `panic` otherwise - with the message here. otherwise we could silently change some other type for `string`, potentially.

docgen/parser/parser.go
```diff
@@ -155,14 +155,63 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
...
+			Name:    nestedTypeName,
+			Comment: nonNilComment(comments[fmt.Sprintf("%s.%d.%d", path, fieldFlag, pos)]),
+			Type: &Type{
+				Name:          fmt.Sprintf("map[%s]%s", keyType, getType(n.GetField()[1]).Name),
+				QualifiedName: "",
+				IsMessage:     false,
+				IsArray:       true,
+				IsEnum:        false,
+			},
+			JSONName: "",
```
mvdan: remove this, just like before.

docgen/parser/parser.go
```diff
@@ -155,14 +155,63 @@ func parseValues(path string, comments map[string]*Comment, values []*google_pro
...
+// extractMapFields generates an array of fields that correspond to map type fields for messages, which are stored in
+// the NestedType field of the DescriptorProto struct
+func extractMapFields(message *google_protobuf.DescriptorProto, path string, comments map[string]*Comment) []*Field {
+	fields := make([]*Field, len(message.NestedType))
+	for i, n := range message.GetNestedType() {
+		// a nested type of map will always have two fields for key and value
+		if len(n.GetField()) != 2 {
+			continue
+		}
+		nestedTypeName := strings.TrimSuffix(n.GetName(), "Entry")
```
mvdan: I find this fragile, like the other comment about `Entry`. There's probably a way to do this that doesn't require messing with strings manually, like comparing type pointers.

docgen/parser/parser.go
mvdan: Or just use `getType`?

docgen/parser/parser.go
kofoworola: @mvdan this particular step checks through fields so messageoptions isn't a property of the struct. Also the line checks if the type name (`.package.Name`) is the same as (`.package.<NameAttribute>Entry`). So if the User named the field FooEntry the type name would be FooEntryEntry

docgen/parser/parser.go
kofoworola: ah got it.

docgen/parser/parser.go
kofoworola: I can't use getType because it returns a FieldDescriptorProto_Type type const.

docgen/parser/parser.go
kofoworola: I added a check for MapEntry type above so this logic should hold as long as it is an entry

mvdan:
@kofoworola the tests are failing though. fine to merge when green

kofoworola:
oh ok thanks will fix.

PR #276, opened 2020-01-21
scopegen: add option to use output version 2

favadi:
output version 2 has following changes:

* all custom types ServiceScope/AuthScope removed, as they are used in no where. Consumes are converting them to map[string]string everywhere
* keeping scope names in generated codes


scopegen/generate/json.go
```diff
@@ -2,18 +2,54 @@ package generate
...
 	}
 	return json.NewEncoder(w).Encode(result)
 }
+
+type model struct {
+	Scopes     map[string]string   `json:"scopes"`
+	AuthScopes map[string][]string `json:"auth_scopes"`
+}
+
+// jsonV2 generates mapping between service full method names and its OAuth2 scope in JSON format.
```
rnd: nit: maybe explain what the difference between jsonV1 and jsonV2 format on the doc here.
summary of what you have on commit message i think it's ok

scopegen/scopegen.go
```diff
@@ -51,6 +53,11 @@ func (s *scopePlugin) Generate(req *plugin_go.CodeGeneratorRequest) (*plugin_go.
 				}
 			case "out":
 				baseDir = v
+			case "output_version":
+				var err error
+				if outputVersion, err = strconv.Atoi(v); err != nil {
+					return nil, fmt.Errorf("unknown output version: err=%s", err.Error())
```
mgfcmartinez: nit: not `:%w`?

testdata/scripts/scopegen_no_swagger_v2.txt
```diff
@@ -0,0 +1,43 @@
...
+
+// Message comment
+type Message struct {
+	Name string `pb:"1" json:"name"`
+}
+
+type Service interface {
+	GetMessage(Message) Message
+}
+-- all.scopes.json.golden --
```
mgfcmartinez: out of scope for the PR but why do we generate the scope files when the definitions don't contain scopes?

mgfcmartinez:
ah, also `Consumers` in the commit message

scopegen/scopegen.go
favadi: `%w` is too new, we are still support go 1.11.

testdata/scripts/scopegen_no_swagger_v2.txt
favadi: It is clearer/more convenience to output an empty scopegen file than omit it. No file generated could mean many things, an empty file clearly told user that the file has successfully generated.

scopegen/generate/json.go
favadi: I added a new section in README.md.

PR #277, opened 2020-02-10
scopegen: update scopegen output v2

favadi:
scopegen output v2 now supports multiple security schemes properly.

Example:

```yaml
securityDefinitions:
  normal_auth:
    scopes:
      read: grant read access
      write: grant write access
  admin_auth:
     scopes:
      admin: grant access to everything

paths:
  /test:
    get:
      security:
        normal_auth:
          - read
          - write
        admin_auth:
          - admin
```

will generate:

```json
{"scopes":{"read":"grant read access","write":"grant write access","admin":"grant access to everything"},"auth_scopes":{"/test":[["read", "write"], ["admin"]]}
```


Hekewa:
`"auth_scopes":{"/test":[["read", "write"], ["admin"]]}` does this mean that we can define an API to need scopes `(A and B) or C` to grant access?

favadi:
> "auth_scopes":{"/test":[["read", "write"], ["admin"]]} does this mean that we can define an API to need scopes (A and B) or C to grant access?

Yes.

Kunde21:
We're conflating roles with scopes here.  Within the RBAC, admin would be granted read and write scopes.  It's no different from any other role and is separate from the scope generation and enforcement.

favadi:
> We're conflating roles with scopes here. Within the RBAC, admin would be granted read and write scopes.

Maybe admin is a bad names for this example, because it implies access control. But we do have use case for an API that requires either: (scope1 and scope2) or (scope3 and scope4).

The logical AND between scopes is documented in [OpenAPI V2 specs](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md).

> Lists the required security schemes to execute this operation. The object can have multiple security schemes declared in it which are all required (that is, there is a logical AND between the schemes).

The logical OR relation between security schemes are defined in:

> A declaration of which security schemes are applied for the API as a whole. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements).

So at least it is an valid use-case defined in OpenAPI V2 specs.

Kunde21:
Are we attempting to support everything that is possible to define in Swagger/OpenAPI specs or are we implementing OAuth2 scopes?  I guess that's the real question that we need to answer.

https://www.oauth.com/oauth2-servers/scope/defining-scopes

favadi:
> Are we attempting to support everything that is possible to define in Swagger/OpenAPI specs or are we implementing OAuth2 scopes?

The way I see is, we do have an use-case for multiple OAuth2 security schemes (a private project) and the OpenAPI V2 specs allows it, so we are adding supports for it. So to answer the question, we are implementing OAuth2 scopes following the OpenAPI V2 specs, for the use-cases we want.

PR #278, opened 2020-02-11
scopegen: adding Any/All helpers for scopegen output_v2

favadi:


scopegen/generate/go.go
```diff
@@ -65,6 +65,12 @@ scopes:
 	goTmplV2 = template.Must(template.New("").Parse(`// Code generated by "scopegen"; DO NOT EDIT.
 package {{ .Package }}
 
+type ServiceScope struct{}
+
+func NewServiceScope() *ServiceScope {
```
Kunde21: We should need a constructor.  Embedding the type in the service struct definition will take care of it.

favadi: > We should need a constructor.

You meant we don't need the constructor?

> Embedding the type in the service struct definition will take care of it.

Already did that in the other PR.

favadi: I removed the constructor in latest commit, as we used it nowhere.

Kunde21: You're right, that's what I meant.  Thank you.

PR #279, opened 2020-02-19
docgen: add optional support for custom headers

karel-3d:
Some Markdown engines (like BlackFriday, used in Hugo) support custom header IDs -
see https://github.com/russross/blackfriday/wiki/Extensions ,
https://github.com/gohugoio/hugo/blob/b2dcd53e3c0240c4afd21d1818fd180c2d1b9d34/markup/blackfriday/convert.go#L190

This can be later used for custom styling based on method types in CSS,
both in the page itself, or in some automatically generated ToC page (as Hugo can do).

With this PR, with option `custom-header-ids=true` in gunkconfig, gunk will generate

`## Add a user {#method-POST-CreateUser}`

and then links to that can then be styled in CSS like

```
a[href*="#method-POST"] {
  color: blue;
}

```

Added as an option, since it's not a standard Markdown


karel-3d:
I have also added a testcase for this

docgen/parser/model.go
```diff
@@ -71,6 +73,13 @@ type Method struct {
 	Operation *options.Operation
 }
 
+func (m *Method) HeaderID() string {
+	name := "method-"
+	name = name + m.Request.Verb + "-"
+	name = name + m.Name
```
favadi: I wonder if we will ever need to have CSS ruleset that target specific API name. Maybe just `"method-" + m.Name` is enough for our purpose?

favadi:
The approach looks good to me, just wonder if we the API name in class name is necessary. Requested @Hekewa and @Kunde21 to review.

docgen/parser/model.go
karel-3d: we want to have different styles for different method types.

karel-3d: Oh you maybe mean `name = name + m.Request.Verb + "-"` is enough?

That cannot be enough, the header IDs need to be unique in a page. This is not a class name, this is anchor ID

karel-3d:
As an aside bonus, the anchor links will now stay the same even when we change the summary - currently, Hugo-generated anchor links change when we change the summary

docgen/parser/model.go
favadi: oops, I understand it wrong in first review. Clear to me now.

karel-3d:
Related - https://github.com/openbank/openbank/pull/80/files

karel-3d:
changed the headers to lowercase, so it's `#method-post-createuser` instead of `#method-POST-CreateUser`.

Merging

PR #281, opened 2020-02-21
test: add gofumpt to travis, tools, generation

karel-3d:
Also fix all gofumpt errors


go.mod
```diff
@@ -13,17 +13,18 @@ require (
...
 	github.com/pkg/errors v0.8.1 // indirect
-	github.com/rogpeppe/go-internal v1.3.0
+	github.com/rogpeppe/go-internal v1.5.2
 	github.com/shurcooL/go v0.0.0-20190704215121-7189cc372560 // indirect
 	github.com/shurcooL/sanitized_anchor_name v1.0.0 // indirect
 	github.com/stretchr/testify v1.3.0 // indirect
 	golang.org/x/sys v0.0.0-20190626221950-04f50cda93cb // indirect
 	golang.org/x/text v0.3.2 // indirect
-	golang.org/x/tools v0.0.0-20190706070813-72ffa07ba3db
+	golang.org/x/tools v0.0.0-20200129045341-207d3de1faaf
```
karel-3d: this got bumped because of the gofumpt

karel-3d:
It seems that go code, that `gunk` itself generates, does not satisfy `gofumpt` (since it just calls `protoc`).

Now I can add running `gofumpt` on the generated code, but it will be a bit harder.... so I don't do that.

mvdan:
conflicts with yourself :)

.travis.yml
```diff
@@ -17,3 +17,4 @@ install: true
 script:
   - go run main.go download all
   - go test -coverprofile=cover.out -v ./...
+  - go install mvdan.cc/gofumpt && git ls-files '*.go' | grep -v testdata | xargs gofumpt -d -s
```
mvdan: I'm pretty sure `-d` will not error if there are changes. did you test this?

see what we do in openbank-services/.drone.yml

karel-3d:
> gunk (the tool, not the source code) should only follow gofmt, not gofumpt. We choose gofumpt here, but we don't want to force that on others.

well, gunk will just return whatever the underlying tool creates, right. So whatever `protoc-gen-go` generates, we should return without any further reformatting.

So you are right basically :)

.travis.yml
karel-3d: You are correct, it does not error. I tested on Travis now :) - https://travis-ci.org/gunk/gunk/builds/653393558?utm_medium=notification&utm_source=github_status 

OK, I will do the `test -z` way

karel-3d: with the `test-z` it does "error properly" - https://travis-ci.org/gunk/gunk/jobs/653396838?utm_medium=notification&utm_source=github_status . nice, thx

karel-3d:
done and tested, you can merge now

Issue #282, opened 2020-02-21
docgen misunderstands basePath swagger option


karel-3d:
Swagger has a `basePath` option, documented here

https://swagger.io/docs/specification/2-0/api-host-and-base-path/

We are using it wrong, at least in the testing files and docgen.

If we have gunk file like this - from `testdata/scripts/extract.txt`:

```
// +gunk openapiv2.Swagger{
//         Swagger: "2.0",
//         Info: openapiv2.Info{
//                 Title:       "Accounts API",
//                 Description: "Provides CRUD operations on the accounts resource.",
//                 Version:     "1.0.0",
//         },
//         Host: "https://openbank.com",
//         BasePath: "/v1",

...[skipped lines]...

type AccountService interface {
        // GetAccount retrieves the detail of an account, selected by its id.
        //
        // +gunk http.Match{
        //         Method: "GET",
        //         Path:   "/v1/accounts/{AccountID}",
        // }
```

the URL is `https://openbank.com/v1/v1/accounts/{AccountID}`.

That is, the `v1` is repeated twice. Once from the URL path of the match, once from the basePath.

It's kind-of documented here

https://github.com/grpc-ecosystem/grpc-gateway/blob/a084954bffbbb1f3a2b2e7dd9586f97dd0a71fd5/protoc-gen-swagger/options/openapiv2.proto#L19-L25

> `base_path` is the common prefix path used on all API endpoints (ie. /api, /v1, etc.). By adding this,
> it allows you to remove this portion from the path endpoints in your Swagger file making them easier
> to read. Note that using `base_path` does not change the endpoint paths that are generated in the resulting 
> Swagger file. If you wish to use `base_path` with relatively generated Swagger paths, the 
> `base_path` prefix must be manually removed from your `google.api.http` paths and your code changed to 
> serve the API from the `base_path`.

also see

https://github.com/grpc-ecosystem/grpc-gateway/issues/918

You can try it at https://editor.swagger.io/ : paste this simplified version, and you can see `/v1` twice on the right in the generated URL

```yaml
swagger: '2.0'
info:
  title: Accounts API
  version: 1.0.0
host: "openbank.com"  
basePath: /v1
schemes:
  - https
paths:
  /v1/accounts:
    get:
      summary: List all accounts
      description: Listing
      operationId: GetAccounts
      responses:
        '200':
          description: Request executed successfully.
          schema: {}
```

[...]

PR #283, opened 2020-02-21
docgen: fix mishandling of basePath and host

karel-3d:
Fixes https://github.com/gunk/gunk/issues/282


docgen/parser/model.go
```diff
@@ -23,6 +23,33 @@ type File struct {
...
+func (f *File) SwaggerScheme() string {
+	if len(f.Swagger.Schemes) == 0 {
+		return ""
+	}
+	hasScheme := map[options.Swagger_SwaggerScheme]bool{}
+	for _, scheme := range f.Swagger.Schemes {
+		hasScheme[scheme] = true
+	}
+	if hasScheme[options.Swagger_HTTPS] {
+		return "https://"
```
favadi: I feel like we could replace the `hasScheme` map with a switch case and always returns `fmt.Sprintf("://", strings.lowercase(options.Swagger_SwaggerScheme.String()))`.

karel-3d: it's an array though

docgen/parser/model.go
```diff
@@ -23,6 +23,33 @@ type File struct {
...
+	}
+	if hasScheme[options.Swagger_HTTP] {
+		return "http://"
+	}
+	// TODO - should we include websocket at all? CURL won't work anyway
```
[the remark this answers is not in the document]
karel-3d: yeah, it is. If by some chance the scheme is only `ws` or `wss`, we will return `curl -S wss://hostname/path/{something}` as an example, even when `curl` itself does not support websocket.

docgen/parser/model.go
karel-3d: that is, `f.Swagger.Schemes` is an array of more schemes, which can be `http`, `https`, `ws`, `wss`, or "unknown". 

This function is used only in `example` in the template with the `curl`, I think.

testdata/scripts/convert_complete_swagger.txt
```diff
@@ -174,7 +174,7 @@ service Service {
 //                 Version: "1.0",
 //         },
 //         Host:     "brank.as",
-//         BasePath: "/v1",
+//         BasePath: "/path",
 //         Schemes: []openapiv2.SwaggerScheme{
 //                 openapiv2.HTTP,
 //                 openapiv2.HTTPS,
```
karel-3d: you can see it here - it can have more schemes, but we want to print `https` in the example if we can?

mvdan:
Merge conflicts :) just re-generate those assets.

mvdan:
The commit message should be telling me what you're fixing and why. The original issue has a lot of context, so I'd like to get a summary in the git history.

karel-3d:
I copied the issue text into commit.

mvdan:
I hate to nitpick, but the commit message is now far too long :) markdown doesn't belong in commit messages either, in general.

It's not like you should only have a few lines, but you literally went from two lines to two pages. You don't really need to quote entire paragraphs from third party docs, or include full YAML examples.

karel-3d:
I just copied the issue haha :D I did not write now. OK I will make it shorter no problem

karel-3d:
ok, now it's great middle ground. :)

karel-3d:
Fixing the failing tests (it's conflict with https://github.com/gunk/gunk/pull/286 )

mvdan:
You forgot to add the "fixes" line at the end of the commit message. Otherwise the closing won't work.

karel-3d:
> You forgot to add the "fixes" line at the end of the commit message. Otherwise the closing won't work.

I thought it's enough that it's here in the PR description

mvdan:
The PR description is lost entirely. Only the commit message stays in master. If I'm doing a `git log`, I want to see what issue you fixed. This would be true even if GitHub's magic with Jira still worked here.

karel-3d:
Added the line to commit message

PR #284, opened 2020-02-24
protoc-gen-go: add json_tag_postproc option

favadi:
Given following gunk code snippet

    // Person represents a homo sapien instance.
    type Person struct {
    	FirstName string `pb:"1" json:"first_name"`
    	LastName string `pb:"2" json:"last_name"`
    }

Without `json_tag_postproc` enabled, the output is:

    	FirstName            string   `protobuf:"bytes,1,opt,name=FirstName,json=first_name,proto3" json:"FirstName,omitempty"`
    	LastName             string   `protobuf:"bytes,2,opt,name=LastName,json=last_name,proto3" json:"LastName,omitempty"`

With `json_tag_postproc` enabled, the JSON struct tag is fixed to has same value as the the one defined in protobuf tag

    	FirstName            string   `protobuf:"bytes,1,opt,name=FirstName,json=first_name,proto3" json:"first_name,omitempty"`
    	LastName             string   `protobuf:"bytes,2,opt,name=LastName,json=last_name,proto3" json:"last_name,omitempty"`

This option is disabled by default.

--

Context: the json tag inconsistency is caused by protoc-gen-go:

* the json struct tag is always the field name: https://github.com/golang/protobuf/blob/d23c5127dc24889085f8ccea5c9d560a57a879d8/protoc-gen-go/generator/generator.go#L2242
* the json attribute inside protobuf struct tag is set to JSOName field and fall back to field name if not defined: https://github.com/golang/protobuf/blob/d23c5127dc24889085f8ccea5c9d560a57a879d8/protoc-gen-go/generator/generator.go#L1555

The post processor fix the inconsistency, the same approach will be used for docgen and swaggergen to fix the wrong json key value in generated examples.

To be discussed:

* this PR added a `json_tag_postproc`  parameter, but it could be a annotation inside gunk file itself.
* the post processor is done by dumb regex search and replace. It could be done with AST parser, but I'm not sure if it is necessary.

### DIFF
--- a/generate/post_process.go
+++ b/generate/post_process.go
@@ -1,0 +1,42 @@
+package generate
+
+import (
+	"strconv"
+
+	"github.com/gunk/gunk/config"
+)
+
+// parseBoolParam returns true if the given parameter name exists and is a truthful value.
+func parseBoolParam(paramName string, gen config.Generator) (bool, error) {
+	for _, param := range gen.Params {
+		if param.Key == paramName {
+			result, err := strconv.ParseBool(param.Value)
+			if err != nil {
+				return false, err
+			}
+			return result, nil
+		}
+	}
+	return false, nil
+}
+
+// postProcess processes the input file before writing to output file.
+func postProcess(input []byte, gen config.Generator) ([]byte, error) {
+	const (
+		protocGenGoCmd           = "protoc-gen-go"
+		jsonTagPostprocParamName = "json_tag_postproc"
+	)
+
+	switch gen.Command {
+	case protocGenGoCmd:
+		jsonTagPostprocEnabled, err := parseBoolParam(jsonTagPostprocParamName, gen)
+		if err != nil {
+			return nil, err
+		}
+		if jsonTagPostprocEnabled {
+			return jsonTagPostProcessor(input)
+		}
+	}
+
+	return input, nil
+}
--- a/generate/post_process_json.go
+++ b/generate/post_process_json.go
@@ -1,0 +1,88 @@
+package generate
+
+import (
+	"bytes"
+	"go/ast"
+	"go/format"
+	"go/parser"
+	"go/token"
+	"reflect"
+	"strings"
+)
+
+// jsonNameFromProtobufTag parse json name from protobuf tag, returns empty string if no json name is defined.
+// example: protobuf:"bytes,1,opt,name=FirstName,json=first_name,proto3"
+// result: first_name
+func jsonNameFromProtobufTag(tag string) string {
+	const jsonPartName = "json="
+
+	if len(tag) == 0 {
+		return ""
+	}
+
+	parts := strings.Split(tag, ",")
+	for _, part := range parts {
+		if strings.HasPrefix(part, jsonPartName) {
+			return strings.TrimPrefix(part, jsonPartName)
+		}
+	}
+	return ""
+}
+
+// jsonNameFromJSONTag returns the json name from json tag, returns empty string if not defined.
+func jsonNameFromJSONTag(tag string) string {
+	if len(tag) == 0 {
+		return ""
+	}
+
+	parts := strings.Split(tag, ",")
+	if len(parts) == 0 {
+		return ""
+	}
+	return parts[0]
+}
+
+func jsonTagPostProcessor(input []byte) ([]byte, error) {
+	const (
+		jsonTagName     = "json"
+		protobufTagName = "protobuf"
+	)
+
+	fset := token.NewFileSet()
+	f, err := parser.ParseFile(fset, "", input, parser.ParseComments)
+	if err != nil {
+		return nil, err
+	}
+
+	ast.Inspect(f, func(node ast.Node) bool {
+		structDecl, ok := node.(*ast.StructType)
+		if !ok {
+			return true
+		}
+
+		for i, field := range structDecl.Fields.List {
+			tagValue := field.Tag.Value
+			tagValue = strings.Trim(tagValue, "`")
+			tag := reflect.StructTag(tagValue)
+
+			jsonName := jsonNameFromJSONTag(tag.Get(jsonTagName))
+			protobufJSONName := jsonNameFromProtobufTag(tag.Get(protobufTagName))
+
+			if jsonName != "" && protobufJSONName != "" && jsonName != protobufJSONName {
+				text := strings.Replace(field.Tag.Value, `json:"`+jsonName, `json:"`+protobufJSONName, 1)
+				structDecl.Fields.List[i].Tag = &ast.BasicLit{
+					ValuePos: field.Tag.Pos(),
+					Kind:     field.Tag.Kind,
+					Value:    text,
+				}
+			}
+		}
+		return true
+	})
+
+	var output bytes.Buffer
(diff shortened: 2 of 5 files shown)


karel-3d:
> This option is disabled by default.

Why?

generate/post_process_json.go
```diff
@@ -0,0 +1,24 @@
+package generate
+
+import (
+	"bufio"
+	"bytes"
+	"regexp"
+)
+
+// jsonTagRegexp is the regular express to make the json struct tag to the same name as the one defined in protobuf tag.
+// Given this input: `protobuf:"bytes,1,opt,name=FirstName,json=first_name,proto3" json:"first_name,omitempty"`
```
karel-3d: nitpick: should be FirstName in the input in the json tag

generate/post_process_json.go
```diff
@@ -0,0 +1,24 @@
...
+// The capture groups are:
+// * $1: `protobuf:"bytes,1,opt,name=FirstName,json=first_name,proto3" json:"
+// * $2: first_name
+// * $3: FirstName
+// * $4: ,omitempty"`
+var jsonTagRegexp = regexp.MustCompile("(`" + `protobuf:"bytes,\d+,opt,name=[\w_]+,json=([\w_]+),proto3" json:")([\w+]+)(,omitempty"` + "`)")
```
karel-3d: just a thought - maybe using https://pkg.go.dev/github.com/fatih/structtag?tab=doc#Parse is more robust and maintanable?

generate/post_process_json.go
favadi: fixed.

generate/post_process_json.go
favadi: maybe, @mvdan your thought?

favadi:
> This option is disabled by default.

to keep backward compatible, we could turn it on by default later.

testdata/scripts/generate_go_json_tag_postproc_disabled.txt
```diff
@@ -0,0 +1,110 @@
+gunk generate -v echo.gunk
```
mvdan: why not use a single script with two subdirectories, and one single `gunk generate ./...` call? if we had one `_disabled` and `_enabled` script for each feature, we'd have twice as many scripts.

Also, you don't need to compare the entire output to a golden file. You could grep for specific things you want or don't want, like `grep 'protobuf:".*whatever"' output.go`. With that method, you could probably avoid 80 lines in each test case.

generate/post_process_json_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestJSONTagPostProcessor(t *testing.T) {
+	tests := []struct {
+		input  []byte
+		output []byte
```
mvdan: make these `string`, and then each case will be simpler - you have N cases, but only one loop body

generate/post_process_json.go
mvdan: Yes, using regex here is... not good.

The simplest alternative would be https://golang.org/pkg/reflect/#StructTag. Note that it doesn't support splitting the value into many comma-separated values, but that's trivial to do via `strings.Split`. I don't think we need to go as far as an external dependency for that.

generate/post_process.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"strconv"
+
+	"github.com/gunk/gunk/config"
+)
+
+func parseJSONTagPostprocParam(gen config.Generator) (bool, error) {
+	const jsonTagPostprocParamName = "json_tag_postproc"
+	for _, param := range gen.Params {
+		if param.Key == jsonTagPostprocParamName {
```
mvdan: do we really not have a func to do this already? we have multiple of these boolean flags in the config file.

favadi:
> I'm also not a big fan of the post_process_json files, each at 20-40 lines. Why not keep it as part of an existing file? I think a new file is only reasonable once you have 100+ lines, at the very least.

after switching to AST, the file is now > 100 lines :D. But please tell if you still want to merge it generate.go, I have no strong opinion on this.

favadi:
@karel-3d @mvdan PR updated, please review.

generate/post_process_json.go
```diff
@@ -0,0 +1,113 @@
...
+			continue
+		}
+
+		// not a struct, skip
+		structDecl, ok := typeSpec.Type.(*ast.StructType)
+		if !ok {
+			continue
+		}
+
+		for _, field := range structDecl.Fields.List {
```
mvdan: You can replace all of this with `ast.Inspect`, and then filtering by the `*ast.StructType` type.

generate/post_process_json.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+		}
+	}
+
+	output := make([]byte, len(input))
+	copy(output, input)
+	for i := range textAreas {
+		area := textAreas[len(textAreas)-i-1]
+		output = append(output[:area.pos], append([]byte(area.text), output[area.end-1:]...)...)
+	}
```
mvdan: This seems fine, but it's a bit weird. It would be simpler to modify the tags in-place in the AST, and then print out the file again. For example, it would be nearly impossible to make "off by one" errors with that method, and it would be much harder to end up with invalid syntax.

generate/post_process_json_test.go
```diff
@@ -0,0 +1,68 @@
...
+	tests := []struct {
+		input  string
+		output string
+	}{
+		{
+			input: `
+package test
+
+// Person represents a homo sapien instance.
+type Person struct {
```
mvdan: nit: this seems the same for every case, so you could just have the field list as the input, and add the rest in the loop.

generate/post_process_json_test.go
```diff
@@ -0,0 +1,68 @@
...
+`,
+		},
+	}
+
+	for _, tc := range tests {
+		output, err := jsonTagPostProcessor([]byte(tc.input))
+		if err != nil {
+			t.Fatalf("failed to postproc json struct tag: err=%s", err.Error())
+		}
+		if bytes.Compare(output, []byte(tc.output)) != 0 {
```
mvdan: simpler if you do `if string(output) != tc.output {`

generate/post_process_json.go
mvdan: Also note that with `ast.Inspect` you can modify nodes in-place, but you can't easily replace them with a different node altogether. You could use https://godoc.org/golang.org/x/tools/go/ast/astutil#Apply for that. I don't think you need to replace entire nodes here, but FYI in case you need to in the future.

generate/post_process_json.go
favadi: replaced the loop with ast.Inspect, that looks so much better. Thank for suggestion.

generate/post_process_json.go
favadi: Updated rewriting the tag directly in ast.File.

generate/post_process_json_test.go
favadi: The example codes is different in second test case.

generate/post_process_json_test.go
favadi: Did that in latest commit.

PR #286, opened 2020-02-25
docgen: fix wrong parameter name in query and response

favadi:
Before this commit, the query and response parameters are generated from field name, changing it to use JSONName instead.


karel-3d:
LGTM

testdata/scripts/extract.txt
```diff
@@ -597,47 +597,47 @@ curl -X POST \
...
-| Name                    | Type   | Description                                                                |
-|-------------------------|--------|----------------------------------------------------------------------------|
-| AccountID               | string | AccountID is the unique identifier of the newly created account.           |
-| CreditTransactionNumber | string | CreditTransactionNumber is the transaction number of the credited account. |
-| DebitTransactionNumber  | string | DebitTransactionNumber is the transaction number of the debited account.   |
-| Minor                   | string | TODO: add comment.                                                         |
+| Name       | Type   | Description                                                                |
+|------------|--------|----------------------------------------------------------------------------|
+| account_id | string | AccountID is the unique identifier of the newly created account.           |
+| number     | string | CreditTransactionNumber is the transaction number of the credited account. |
```
karel-3d: Sorry for saying it just now, but - this does not seem correct. This looks like some bug, no?

testdata/scripts/extract.txt
```diff
@@ -597,47 +597,47 @@ curl -X POST \
...
+| number     | string | DebitTransactionNumber is the transaction number of the debited account.   |
```
karel-3d: since there are two fields named `number`, while originally there is `CreditTransactionNumber` and `DebitTransactionNumber`. No idea where it goes from, probably unrelated to this PR though

karel-3d: I tracked the issue to `pulpMd`; pulpMd is too dense to me to get what exactly is going on there, so I made an issue :)

https://github.com/Kunde21/pulpMd/issues/5

PR #288, opened 2020-03-03
docgen: update pulpMd (+ transitively, markdownfmt) to fix markdown bugs

karel-3d:
I have put pulpMd and markdownfmt to better shape (and added some tests there); now let's use the new versions.

Also fixed invisible space in annex template

Also fixed master test failure, related to recent docgen changes


karel-3d:
For some reason, I cannot add reviewers here, so /cc @favadi

docgen/templates/annex.md
```diff
@@ -2,7 +2,7 @@
 
 {{range $m := .Messages}}
 
-## {{$m.Name}}
+## {{$m.Name}}
```
karel-3d: these are tiny changes, that should be a separate commits; but I didn't want to make 2 commits, because changing the test scripts is so painful every time :/

testdata/scripts/extract.txt
```diff
@@ -552,15 +552,17 @@ msgstr ""
...
 msgstr ""
 -- all.md.golden --
-# Accounts API v1.0.0
+Accounts API v1.0.0
+===================
 
 Provides CRUD operations on the accounts resource.
 
-* Host `https://openbank.com`
+*   Host `https://openbank.com`
```
karel-3d: markdownfmt now uses more spaces, because that is sometimes necessary by blackfriday (a kind-of buggy markdown parser, that we use in both markdownfmt and hugo)

testdata/scripts/extract.txt
```diff
@@ -632,12 +635,12 @@ curl -X POST \
...
-| Name       | Type   | Description                                                                |
-|------------|--------|----------------------------------------------------------------------------|
-| account_id | string | AccountID is the unique identifier of the newly created account.           |
-| number     | string | CreditTransactionNumber is the transaction number of the credited account. |
-| number     | string | DebitTransactionNumber is the transaction number of the debited account.   |
-| minor      | string | TODO: add comment.                                                         |
+| Name                      | Type   | Description                                                                |
+|---------------------------|--------|----------------------------------------------------------------------------|
+| account_id                | string | AccountID is the unique identifier of the newly created account.           |
+| credit_transaction_number | string | CreditTransactionNumber is the transaction number of the credited account. |
```
karel-3d: this was an actual bug in markdownfmt

testdata/scripts/extract.txt
favadi: that looks weird, do you know why blackfriday requires double spaces here?

testdata/scripts/extract.txt
```diff
@@ -690,7 +696,8 @@ curl -X DELETE \
 | 403    | Returned when the user does not have permission to access the resource.                |
 | 500    | Returned when an unexpected error occured on the server side.                          |
 
-## Retrieve an account
+Retrieve an account
+-------------------
```
favadi: why it changes from ## to ______?

testdata/scripts/extract.txt
karel-3d: Blackfriday is not commonmark compliant in weird ways. One of the weirdest issues is the spacing of lists

https://github.com/russross/blackfriday/issues/329 (3 years old issue now)

This does not parse correctly
```
* foo
  * bar
    * baz
```

this does
```
*   foo
    *   bar
        *   baz
```

this does not parse correctly
```
1. A paragraph

   item continues
```

this does (1 extra space)

```
1.  A paragraph

    item continues
```

karel-3d: Hugo switched to GoldMark some time ago, but not the version we are using.

testdata/scripts/extract.txt
karel-3d: Because I thought it's more "common" way to write headers in markdown, so markdownfmt converts `## foo` to 

```
foo
---
```

(original markdownfmt did that too, just our fork didn't)

karel-3d:
can you please merge when you think it's OK? I don't have write access here :(

favadi:
I would like to have @Kunde21 take a look at this before merging.

karel-3d:
btw I just fixed markdownfmt+pulpMd to use goldmark instead of blackfriday... our (old) hugo version still uses blackfriday though, so I still format lists with this wider spacing

assets/generated/annex.md
```diff
@@ -2,7 +2,7 @@
 
 {{range $m := .Messages}}
 
-## {{$m.Name}}
+## {{$m.Name}}
```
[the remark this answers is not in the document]
karel-3d: The original file had an invisible space (`&nbsp;`), probably as a mistake, This is a change to a normal space.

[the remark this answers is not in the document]
mvdan: SGTM. The commit message should probably clarify this.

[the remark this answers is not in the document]
karel-3d: I *can* break it into more smaller commits if needed, I was just too lazy to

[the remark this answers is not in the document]
mvdan: No need, I think. It's just that, seeing the diff in plaintext, this change is confusing.

karel-3d:
Updated to new pulpMd version

Also update commit message to clarify the comments here (the markdown changes + the invisible space)

testdata/scripts/extract.txt
karel-3d: I now switched our Hugo to the version with Goldmark, so I guess I can change it back so it uses normal two spaces :)

karel-3d: I switched to Hugo everywhere in internal docs, so this is done now

karel-3d:
finally!

@favadi maybe this messes up your PRs elsewhere, sorry for that, but finally merging

PR #289, opened 2020-03-04
generate_complete_swagger: wrong expected value for base path

favadi:
This failure happens because @karel-3d PR https://github.com/gunk/gunk/pull/283 got merged before https://github.com/gunk/gunk/pull/287. Both PRs CI passed, but master is failing.


karel-3d:
I corrected it in https://github.com/gunk/gunk/pull/288 but it's OK to fix it first so we have a green master :)

PR #290, opened 2020-03-04
json_tag_postproc: fix panic with gRPC plugin enabled

favadi:
When post process json tag, skipping struct field with no tag defined.

### DIFF
--- a/testdata/scripts/generate_go_json_tag_postproc.txt
+++ b/testdata/scripts/generate_go_json_tag_postproc.txt
@@ -7,6 +7,7 @@ module testdata.tld/util
 
 -- enabled/.gunkconfig --
 [generate go]
+plugins=grpc
 json_tag_postproc=true
 
 -- disabled/.gunkconfig --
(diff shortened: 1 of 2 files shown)


favadi:
requesting review from @karel-3d

karel-3d:
What does `plugins=grpc` actually do?

The outputs are the same whether I put it in `.gunkconfig` or not 😃except for the panic

testdata/scripts/generate_go_json_tag_postproc.txt
```diff
@@ -21,6 +22,11 @@ type Person struct {
 	LastName string `pb:"2" json:"last_name"`
 }
 
+// AnExampleService is just an example.
```
karel-3d: this should be both in enabled and disabled to test the grpc, right?

karel-3d: now it's just in disabled (so it never tests the panic handling)

karel-3d:
please copy the `AnExampleService` to enabled too. (Or, just have it once and then do `cp` in the testscript first.)

PR #291, opened 2020-03-04
generate: use a separate lock file when downloading protoc

mvdan:
Before this patch 'sudo rm -rf .cache && go test ./...' would result in
lots of errors like:

	error: error executing "[...]/protoc-v3.9.1": [...]: text file busy

This was because we were using the binary itself as the file being
locked. Because the lock was grabbed with write access, and many
operating systems like Linux forbid executing a binary that's open for
writing, we were seeing data races when downloading.

Here's a description of the chain of events that could lead to a race
between two tests, assuming the empty cache:

	Test A: finds protoc is missing, locks the binary for writing
	Test B: finds protoc is missing, locks the binary for writing
	Test A: wins the race for the lock, begins downloading protoc
	Test B: loses the race for the lock, blocks
	Test A: finishes the download, closes the file, and releases the lock
	Test B: unblocks and grabs the lock, opening the file for writing
	Test A: tries to verify that protoc works, executing the binary
	Test A: errors, since Test B has the file open for writing

There are other cases where we could hit such an error, too. The above
is just one of the examples. To fix this problem, use a separate lock
file, as explained in the added comments.

The tests now reliably pass on Go 1.14, regardless of whether the cache
file exists or not.

### DIFF
--- a/generate/download.go
+++ b/generate/download.go
@@ -24,6 +24,9 @@ const defaultProtocVersion = "v3.9.1"
 // If the version is not specified, the latest version is fetched from GitHub.
 // If both version and path are specified and a file already exists at the path,
 // it checks whether the output of `protoc --version` is an exact match.
+//
+// Note that this code is safe for concurrent use between multiple goroutines or
+// processes, since it uses a lock file on disk.
 func CheckOrDownloadProtoc(path, version string) (string, error) {
 	if version == "" {
 		version = defaultProtocVersion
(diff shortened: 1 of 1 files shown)


generate/download.go
```diff
@@ -45,10 +48,21 @@ func CheckOrDownloadProtoc(path, version string) (string, error) {
...
+	// directly as the lock can lead to "text file busy" errors.
+	unlock, err := lockedfile.MutexAt(dstPath + ".lock").Lock()
+	if err != nil {
+		panic(err)
+	}
+	defer unlock()
+	// We are the only goroutine with access to dstPath. Check if it already
+	// exists. Using lockedfile.OpenFile allows us to do an atomic write
+	// when it doesn't exist yet.
 	dstFile, err := lockedfile.OpenFile(dstPath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0775)
```
favadi: my understanding is we already using a lock file mutex above, so lockedfile.OpenFile could be replaced with os.OpenFile without any ill effect. Is this correct?

mvdan: Not exactly, because we still want the file write to be atomic. This is what the comment just above explains. For example, if the download fails, or if the user hits ^C while we are downloading or writing the file, or if the power shuts off mid-process, we don't want to leave a broken protoc binary.

favadi: right, sorry I didn't read the comment carefully.

PR #293, opened 2020-03-11
docgen: extract enums from dependencies into annex

karel-3d:
currently, we extract only enums from the same file, not from dependencies. See the new test.

### DIFF
--- a/testdata/scripts/extract_nested_enum.txt
+++ b/testdata/scripts/extract_nested_enum.txt
@@ -1,0 +1,297 @@
+# note that Account's MajorType is in a different package "types",
+# but still generated in the transactions/all.md annex
+# (and also types/types.md, but we don't check that here)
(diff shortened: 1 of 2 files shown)


karel-3d:
requesting @kofoworola too

testdata/scripts/extract_nested_enum.txt
```diff
@@ -0,0 +1,300 @@
+# the two main points to note in this test are that the Transaction struct's Amount
```
karel-3d: oh I should change this

karel-3d:
note that this and #294 clash - if both are merged at same time, tests will fail

karel-3d:
oh s**t, now it adds *all* the enums from the types, not just the one used in the main file, to the annex. I will look on how to fix that.

PR #294, opened 2020-03-11
docgen: better enum descriptions

karel-3d:
Currently, the enum description includes the enum type repeated. Let's remove them.

See testscript changes


karel-3d:
requesting @kofoworola too

note that this and https://github.com/gunk/gunk/pull/293 clash - if both are merged at same time, tests will fail

karel-3d:
Also, if I am not wrong, the JSON GRPC needs to receive the enum value as a string without the type prefix, so this is actually more correct.

PR #296, opened 2020-03-11
docgen: Add custom header IDs to responses, objects, enums

karel-3d:
I am doing javascript magic on the frontend for collapsing enums and nested messages;
it needs to have the same header IDs, even if we later change the text.

So let's make the header IDs fixed

Note that this is only triggered with `custom-header-ids=true`, see testscripts

Moved the config boolean away from model and using a function instead, as
it's probably more readable like this (also the File object is not accessible from sub-templates)


assets/generated/api.md
```diff
@@ -8,11 +8,7 @@
 {{- range $s := .Services}}
 {{- range $m := $s.Methods}}
 
-{{if $.CustomHeaderIds}}
-## {{GetText $m.Operation.Summary}} {#{{$m.HeaderID}}}
-{{else}}
-## {{GetText $m.Operation.Summary}}
-{{end}}
+## {{GetText $m.Operation.Summary}} {{CustomHeaderId $m.HeaderID}}
```
favadi: will it add a trailing space if `CustomHeaderId` returns empty string? Compare to current output?

karel-3d: it will, but `gofmt` removes it again so the tests work

PR #300, opened 2020-03-12
docgen: do not generate "annex" (no services file)

karel-3d:
Currently, we generate "annex file" for gunk files without services

This file is, however, basically useless; nobody wants to read all sub-types without context, and all
the messages should be already documented in the services.

And if we have any types, not "reachable" from services, they should not be generated at all anyway.

I have changed some of the tests that depended on Annex and slightly simplify them.

Note that with this change, we now don't need Messages in the File parse model at all!

Also note that this uncovered some bug in the JSON example generation for maps.

(note: please do not merge before we try this in the whole pipeline)


testdata/scripts/extract_import.txt
```diff
@@ -38,44 +71,60 @@ type Amount struct {
...
+
+##### Objects
+
+###### Amount
 
 | Name | Type   | Description                        |
 |------|--------|------------------------------------|
-| Cur  | string | Cur is the currency of the amount. |
-| Num  | string | Num is the value of the amount.    |
+| cur  | string | Cur is the currency of the amount. |
```
karel-3d: note that it incidentally fixes the JSON names here

docgen/parser/model.go
```diff
@@ -16,7 +17,6 @@ type FileDescWrapper struct {
 // File is a proto parsed file.
 type File struct {
 	Swagger  *options.Swagger
-	Messages map[string]*Message
```
karel-3d: this is not needed anymore in the new doc (since the messages are in Services)

enums are still needed, because they are in doc page annex; I will remove them in a later diff

docgen/parser/model.go
```diff
@@ -53,6 +53,33 @@ func (f *File) HasServices() bool {
 	return len(f.Services) > 0
 }
 
+func (f *File) CheckSwagger() error {
```
karel-3d: these are necessary so that the markdown result builds, and does not have empty headers (which breaks FE)

testdata/scripts/generate_docgen_map.txt
```diff
@@ -28,28 +45,71 @@ type Price struct {
...
-| Name           | Type             | Description                                       |
-|----------------|------------------|---------------------------------------------------|
-| MapStringInt   | map[string]int32 | MapStringInt is a simple map of a <string,int>.   |
-| MapIntString   | map[int]string   | MapStringInt is a simple map of a <int,string>.   |
-| MapStringPrice | map[string]Price | MapStringPrice is a sample map of <string,price>. |
+#### Response body
+
+| Name             | Type                                              | Description |
+|------------------|---------------------------------------------------|-------------|
+| map[string]int32 | MapStringInt is a simple map of a <string,int>.   |             |
```
karel-3d: this looks like a bug

testdata/scripts/generate_docgen_map.txt
```diff
@@ -28,28 +45,71 @@ type Price struct {
...
+| map[int]string   | MapStringInt is a simple map of a <int,string>.   |             |
+| map[string]Price | MapStringPrice is a sample map of <string,price>. |             |
+
+Example:
+
+```json
```
karel-3d: this looks like a bug

testdata/scripts/generate_docgen_map.txt
```diff
@@ -28,28 +45,71 @@ type Price struct {
...
+{
+  "": "map[string]int32",
```
karel-3d: this looks like a big

karel-3d:
I don't understand why travis fails.

karel-3d:
Oh crap, we still need the annex pages for errors, damn it

Since as currently defined, gunk interfaces / grpc services do not have errors defined

```
type ATMService interface {
    GetATMs() GetATMsResponse // not (GetATMsResponse, APIError)
}
```

so `APIError` is technically not "reachable" from any message, since it's never returned from the interface - but might be actually returned from the JSON API?

I am not exactly sure how does the JSON GRPC Gateway work with errors and where is the error returned. Hmmmm. What to do with it...

testdata/scripts/generate_docgen_map.txt
favadi: My guess is an unclosed JSON tag?

karel-3d: nope, the tag is fine

docgen/generate/generate.go
```diff
@@ -25,10 +27,21 @@ func Run(w io.Writer, f *parser.FileDescWrapper, lang []string, customHeaderIds
...
 
-	tplName := "annex.md"
-	if api.HasServices() {
-		tplName = "api.md"
+	if !api.HasServices() {
+		return nil, ErrNoServices
+	}
+
+	err = api.CheckSwagger()
+	if err != nil {
```
mvdan: `if err := f(); err != nil {`

docgen/generate/generate.go
```diff
@@ -25,10 +27,21 @@ func Run(w io.Writer, f *parser.FileDescWrapper, lang []string, customHeaderIds
...
+		name := ""
+		if f.Name != nil {
+			name = *f.Name
+		}
+		return nil, fmt.Errorf("swagger error in file %s: %s", name, err.Error())
+		// TODO use following after switch to go 1.13
```
mvdan: assign the TODO to yourself?

docgen/parser/model.go
mvdan: please document this in the godoc.

PR #301, opened 2020-03-12
all: rework the use of test caches in test scripts

### LINKED ISSUE #268: test: stop writing to the repo and leaving read-only files behind
After a failed test run, I'm left with unstaged git changes, and a `.cache` directory holding read-only files that I can't `git clean` since they were created by the module cache.

All of this is ugly and needs to be done better.

mvdan:
We were using .cache/home/ under the root of the repository to speed up
incremental test runs, to avoid re-downloading gunk's dependencies at
every test run.

However, this had the major disadvantage of polluting the git checkout,
and leaving read-only files behind, as the Go cache would also end up
there.

Instead, let the tests use the user's real cache directory by default.
Tests which want to use an empty cache set up their own. To simplify
that in a cross-platform way, add a GUNK_CACHE_DIR environment variable.

Overall, the new approach is simpler and tidier.

Fixes #268.

### DIFF
--- a/generate/download.go
+++ b/generate/download.go
@@ -34,10 +34,15 @@ func CheckOrDownloadProtoc(path, version string) (string, error) {
 
 	dstPath := path
 	if dstPath == "" {
-		// Get the os specific cache directory.
+		// Get the OS-specific cache directory.
 		cachePath, err := os.UserCacheDir()
 		if err != nil {
 			return "", err
+		}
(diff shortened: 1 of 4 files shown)


karel-3d:
works for me, `.cache` is still created, but without read-only files

mvdan:
You're right that the directory is still created, I forgot about that. It's for the compiled binaries for the tests. I think that's still OK, as I don't see a clearly better approach.

PR #302, opened 2020-03-13
docgen: move enums from in `api.md` to methods

karel-3d:
This removes the "annex" part of API documentation and moves it to methods
where they are used, not in the back of the file.

This is different from https://github.com/gunk/gunk/pull/300 ;
this does not concern `annex.md`, but `annex` part of `api.md`.


testdata/scripts/extract_append.txt
```diff
@@ -369,18 +360,6 @@ Example:
...
-#### EntityType
-
-EntityType describes the type of the entity.
-
-| Value | Description                      |
-|-------|----------------------------------|
-| Pers  | Pers is an individual customer.  |
-| Org   | Org is an organisation customer. |
-
 Annex
```
karel-3d: this annex is from `annex.md`, since we do append

testdata/scripts/extract_code_sample.txt
```diff
@@ -229,9 +226,6 @@ msgstr ""
 msgid "DueDate is the loan maturity date."
 msgstr ""
 
-msgid "EntityType describes the type of the entity."
```
karel-3d: `EntityType` dropped out of the file completely, since no service is using it

favadi:
The changes LGTM, but please wait for @mvdan's review.

mvdan:
So this is an alternative to #300? Also LGTM. Just pick one and merge it :) I presume this one is preferrable, from looking at the output files.

karel-3d:
@mvdan nope

we have 2 things, one is `annex` in api files, and one is `annex.md`. Two different things. I want to get rid of both.

mvdan:
Ah, I read `this is different` as in `this is a different approach to fix the same problem`, not `this is a different problem to be fixed entirely`. Merging both SGTM.

PR #303, opened 2020-03-18
CI: switch to Actions, test on Go 1.13 and 1.14

mvdan:


mvdan:
Damnit, there is no way to get Actions to pick this up. I'll add a dummy file to master instead.

PR #304, opened 2020-03-18
CI: switch to Actions, test on Go 1.13 and 1.14

mvdan:


mvdan:
Note that I left gofumpt out on purpose. The old check was done a bit weird; it ran on all platforms and Go versions, so that's a waste of time/resources. It should only run on the latest Go version and on Linux.

karel-3d:
> The old check was done a bit weird; it ran on all platforms and Go versions, so that's a waste of time/resources. It should only run on the latest Go version and on Linux.

While true, it's almost instant, no?

Issue #310, opened 2020-04-08
Multi-line comments produce ugly pb.go code and ugly documentation md


karel-3d:
If you have this in .gunk

```
type Foo struct {
    // Num is
    // a number
    Num string `pb:"1" json:"num"`
}
```

the generated md by docgen has `Num isa number`, and the generated all.pb.go has

```
type Amount struct {
    // Num is
    //a number
    Num string `(....all the PB stuff...)`
}
```

The second thing used to be corrected by gofumpt in our pipeline, but it now has some bugs related to adding spacing, it's better to generate it correctly in the first place

karel-3d:
Related in gofumpt - https://github.com/mvdan/gofumpt/issues/52 , https://github.com/mvdan/gofumpt/pull/50 , https://github.com/mvdan/gofumpt/commit/4bdc928868e60b5ac1a986e3565f24c579868f80

PR #311, opened 2020-04-08
Proper leading spaces for multi-line comments

karel-3d:
go's ast.TypeSpec.Doc.Text() trims left-trailing spaces on each line of multi-line comment,
while proto's LeadingComments needs them

block comments still look bad, but that's not a priority now

Fixes https://github.com/gunk/gunk/issues/310

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -575,13 +575,22 @@ func (g *Generator) addDoc(text string, path ...int32) {
 	if text == "" {
 		return
 	}
+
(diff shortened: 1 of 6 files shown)


karel-3d:
Fixes the go part of https://github.com/gunk/gunk/issues/310 ; maybe also the markdown (not sure)

karel-3d:
.. and from the test failures, it seems it actually fixes the markdown too. Good.

PR #312, opened 2020-04-10
Make .pb.go deterministic by deterministic marshal

karel-3d:
We have large diffs in generated .pb.go files even when there is no change in the protobuf itself. Example:

https://github.com/openbank/openbank/commit/1535696c0f58698898b266dee9bbd4ff144dc874#diff-790a2e5fdb0867ef31e92c5b78946f3c

I found out it's actually caused by non-deterministic marshaling of maps; by default, proto.Marshal is non-deterministic with maps, and grpc uses maps to save response types.

If you actually un-gzip the byte array and diff the filedescriptor, the only changes are in the order of the maps in GRPC.

You can marshal deterministcally using proto.Buffer, so I added a helper func that does that

### DIFF
--- a/dump/dump.go
+++ b/dump/dump.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/gunk/gunk/generate"
+	"github.com/gunk/gunk/protoutil"
 )
 
 // Run will generate the FileDescriptorSet for a Gunk package, and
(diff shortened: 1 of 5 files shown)


karel-3d:
@mvdan see https://github.com/golang/protobuf/issues/86 :) 

I don't know

I suspect: sorting can slow thing down, and when using it for communication between processes, the determinism and map order does not matter.

However, we are not using it just for communication, we are using protobuf itself to encode the protobuf definition. And then THAT meta-description is saved in the .pb.go by protoc-gen-go.

karel-3d:
FYI, I made a diff here, we can remove the helper function/package again if it's merged

https://github.com/golang/protobuf/pull/1073

PR #313, opened 2020-07-14
docgen: add only_external option

favadi:
to allow to skip all method not tagged with "external"

### DIFF
--- a/docgen/parser/parser.go
+++ b/docgen/parser/parser.go
@@ -26,14 +26,21 @@ const (
 	enumFlag    = 5
 )
 
+type OnlyExternalOpt int
+
+const (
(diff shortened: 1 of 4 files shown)


docgen/parser/parser.go
```diff
@@ -26,14 +26,21 @@ const (
 	enumFlag    = 5
 )
 
+type OnlyExternalOpt int
```
karel-3d, line 29: advantage over bool?

favadi: The upside is it is clearer on function call: `func DoSomething(OnlyExternalEnabled)` vs `func DoSomething(true)`. The downside is we pollute the package with a new type.

karel-3d: good point

PR #315, opened 2020-11-03
Fix (docgen: hide) cannot hide nested struct fields

uyen-brankas:
When filter out the (docgen: hide) in messages, filter it out in all messages instead of only messages on the top level (with name)

### DIFF
--- a/docgen/parser/parser.go
+++ b/docgen/parser/parser.go
@@ -461,16 +461,17 @@ func generateResponseExample(messages map[string]*Message, name string) (string,
 }
 
 // normalizeRequestFields is used to filter out all fields marked with an predefined indicator of a request message.
-func normalizeRequestFields(messages map[string]*Message, name string) {
+func normalizeRequestFields(messages map[string]*Message) {
 	const hidingIndicator = "docgen: hide"
-	var filtered []*Field
-	for _, f := range messages[name].Fields {
-		if strings.Contains(f.Comment.Leading, hidingIndicator) {
-			continue
+	for _, message := range messages {
+		var filtered []*Field
+		for _, f := range message.Fields {
+			if !strings.Contains(f.Comment.Leading, hidingIndicator) {
+				filtered = append(filtered, f)
+			}
 		}
-		filtered = append(filtered, f)
+		message.Fields = filtered
 	}
-	messages[name].Fields = filtered
 }
 
 func parseRequest(rule *method.HttpRule, messages map[string]*Message, name string) (*Request, error) {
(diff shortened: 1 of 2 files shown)


docgen/parser/parser.go
```diff
@@ -463,14 +463,16 @@ func generateResponseExample(messages map[string]*Message, name string) (string,
...
-	for _, f := range messages[name].Fields {
-		if strings.Contains(f.Comment.Leading, hidingIndicator) {
-			continue
+	for _, message := range messages {
+		var filtered []*Field
+		for _, f := range message.Fields {
+			if strings.Contains(f.Comment.Leading, hidingIndicator) {
+				continue
+			}
+			filtered = append(filtered, f)
```
kofoworola, line 472: For this you could try reformatting as 
```
for _, f := range message.Fields {
			if !strings.Contains(f.Comment.Leading, hidingIndicator) {
				filtered = append(filtered, f)
			}
		}

```
to avoid use of continue, but it is just a suggestion, this is fine too 🙂

testdata/scripts/generate_docgen_only_external.txt
```diff
@@ -53,9 +53,31 @@ type ExampleRequest struct {
     FooName string `pb:"1" json:"fooName"`
     // BarName is just an example (docgen: hide).
     BarName string `pb:"2" json:"barName"`
-    // FooFooName is also should be hiding.
+	// FooFooName is also should be hiding.
```
kofoworola, line 56: nit: format line properly and use proper amount of spaces

testdata/scripts/generate_docgen_only_external.txt
```diff
@@ -53,9 +53,31 @@ type ExampleRequest struct {
...
     // second line: (docgen: hide).
-    FooFooName string `pb:"3" json:"barName"`
+    FooFooName string `pb:"3" json:"fooFooName"`
+	// FooStructInstance has a hidden field.
```
kofoworola, line 59: nit: same as above

kofoworola (changes requested):
LGTM, just need to format test file properly. Thanks for this!

docgen/parser/parser.go
```diff
@@ -463,14 +463,16 @@ func generateResponseExample(messages map[string]*Message, name string) (string,
 // normalizeRequestFields is used to filter out all fields marked with an predefined indicator of a request message.
 func normalizeRequestFields(messages map[string]*Message, name string) {
 	const hidingIndicator = "docgen: hide"
-	var filtered []*Field
-	for _, f := range messages[name].Fields {
-		if strings.Contains(f.Comment.Leading, hidingIndicator) {
-			continue
+	for _, message := range messages {
```
favadi, line 466: with this changes, does that meant that we now loop through all messages , not just the one with matching name? If yes:

* name is now unused
* I think there are duplicated efforts looping through all those messages. Could you try to move normalizeRequestFields to the top level function (ParseFile).

favadi (approved):
I think the fix works, just suggest something can be improved. Also, please explain the reason of the bug and what is the fix in commit message.

PR #316, opened 2021-02-08
Fixes TestGenerate

karelbilek:
Needed two changes in code:

* add full package to proto package path, to shut down protoc-gen-go warning
* add some nonsense Span to satisfy protoc-gen-go sanity checker

Also needed to change the actual test data.

I have also added more error context for better debugging, I left it in.


testdata/generate/all.pb.go
```diff
@@ -1,4 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.25.0
```
karelbilek, line 3: this will probably fail the next time `protoc-gen-go` is updated, but that's ok?

karelbilek:
tests are failing as I just fixed the first failing test, the others are still failing :( 

one by one. if no objections I will merge it soon (I didn't realize I have commit rights here)

PR #326, opened 2021-02-22
openapi fixes to properly handle message opts

Kunde21:
* Fix openapi schema option to apply to messages, in order to add
optional data, such as required fields.
Unknown if there is any code depending on the field handling of
JSONSchema, so that has not been disabled.

* Resolve packages to generate Python code.
Python packages cannot have the common dot '.' separator.  Replace
occurances with a slash '/' for python package resolution.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -202,6 +202,21 @@ func (c configWithBinary) actualCommand() string {
 	return *c.binary
 }
 
+// findPkg resolves package names for languages with different naming requirements and restrictions.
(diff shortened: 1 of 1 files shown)


karelbilek:
merged in separate commits ( https://github.com/gunk/gunk/commit/5fdeea1441bc11b15a7043f53cc1f251d56734f9 , https://github.com/gunk/gunk/commit/bb45fe6e8de75c381fe5a8e953bc8c17d350d19c )

PR #335, opened 2021-04-16
Use openapiv2 example in docgen requests

karelbilek:
Use like this (also see extract_example.txt)

```
// CreateAccountRequest wraps all required fields for an account creation.
//
// +gunk openapiv2.Schema{
//         Example: `{
//             "foo": "bar"
//         }`,
// }
type CreateAccountRequest struct {
```


karelbilek:
Note that the string still needs to be a valid JSON, otherwise another part of the code fails.

PR #336, opened 2021-04-21
add support to add example on field level

dimzst:
value must be a json value (refer to extract_example.txt for more
example)
usage example:
```
// OwnerName is the name of the account's owner.
//
// +gunk openapiv2.Schema{
//         JSONSchema: openapiv2.JSONSchema{
//             Example: `"John Doe"`,
//         },
// }
OwnerName string `pb:"21" json:"owner_name"`
```

### DIFF
--- a/docgen/parser/parser.go
+++ b/docgen/parser/parser.go
@@ -232,12 +232,18 @@ func parseMessages(
 	// convert each proto message to our Message representation
 	for i, m := range messages {
 
-		fixedExample := ""
-		schemaI := proto.GetExtension(m.GetOptions(), options.E_Openapiv2Schema)
-		if schemaI != nil {
-			schema := schemaI.(*options.Schema)
-			if schema != nil {
-				fixedExample = schema.Example
+		var fixedExample map[string]interface{}
+		if proto.HasExtension(m.GetOptions(), options.E_Openapiv2Schema) {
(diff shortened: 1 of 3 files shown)


docgen/docgen.go
```diff
@@ -20,6 +21,7 @@ import (
 const modeAppend = "append"
 
 func main() {
+	log.Println("SOMETHING")
```
karelbilek, line 24: remove this

docgen/parser/parser.go
```diff
@@ -232,12 +232,18 @@ func parseMessages(
 	// convert each proto message to our Message representation
 	for i, m := range messages {
 
-		fixedExample := ""
-		schemaI := proto.GetExtension(m.GetOptions(), options.E_Openapiv2Schema)
-		if schemaI != nil {
-			schema := schemaI.(*options.Schema)
-			if schema != nil {
-				fixedExample = schema.Example
+		var fixedExample map[string]interface{}
```
karelbilek, line 235: What's the point of this change?

karelbilek: Ah I see below

dimzst: Yup, that is to support nested message

karelbilek:
great job

PR #338, opened 2021-07-06
Validate field tags

karelbilek:
I have noticed gunk allows invalid (un-terminated) JSON tags,
also it allows double JSON tags and other (never used) tags.

I have copied over the go vet code to check correct tags.


karelbilek:
Hm, I need to remove the strict check, as we also add `db` tags, because we use the same structs from front-end/GRPC also for SQL and depend on the `db` tags :O 

I guess those can be duplicated though (go vet itself checks just `json` duplications).

PR #339, opened 2021-07-26
Check that field tags only have pb and json; also check correctness

karelbilek:
In our large codebase with gunk, we have introduces some errors by either wrong
tags or by tags that are not parseable.

Add checks for that in gunk, so it does not silently ignore it.

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


karelbilek:
Uh I see that I duplicated some work from this one https://github.com/gunk/gunk/pull/338

Maybe will be better to add it there instead of introducing yet another dependency...

PR #342, opened 2021-08-02
docgen: update auth template

grikwong:
This updates the authorization method of the docgen-generated api
documentation from Bearer token to API Key.


karelbilek:
🤷

PR #347, opened 2021-09-23
Replace space check for comma check

karelbilek:
The go vet check includes check for space after comma, which makes sense in
go vet, as it catches `json:"foo, omitempty"`. However I simplified the code here
and include a bug, where JSON cannot have a space in name.

I have instead changed this to checking for comma, as we do not allow `json:"foo,omitempty"` anyway.


karelbilek:
I have noticed this because a real life code needed space in a JSON key... which is weird but that's life I guess

PR #352, opened 2021-12-22
doc: Add comments and tests

chanbakjsd:
This PR adds comments to functions in `generate` and tests (mainly for `gunk vet`).
It also updates the example gunk file to a newer format.


kenshaw:
Merged.

PR #354, opened 2021-12-24
loader: Use Go Overlays to parse Gunk-only packages

chanbakjsd:
Use Go overlays instead to make the Go toolchain parse Gunk packages. This allows "fake Go files" to be generated even in read-only file systems like the Go module cache.

More details are provided in the commit message.


kenshaw:
Merged.

PR #356, opened 2021-12-27
loader: allow package name mismatch on Go files

chanbakjsd:
This allows compilation to succeed even if the Gunk package name has changed if the generated files are in the same directory.


chanbakjsd:
I was notified that a workflow has failed for this PR.
Inspecting the logs, it seems like it's some internal Git fetching error, please re-run the workflow, thanks.

PR #374, opened 2022-07-11
Fix version checks for protoc 21+, download ARM binary on mac if possible

### LINKED ISSUE #373: protoc has universal binary since 3.20 for macOS
For macOS, protoc since 3.20 has universal binary, which supports ARM.

The downloader magic should be changed to support that.

karelbilek:
Fixes #373


hhhapz:
!remindme 15years when protoc v3.100 releases and this implementation breaks :p

karelbilek:
I fail to see what would break with v3.100?

